PR #43, opened 2016-05-07
Remove line too long for lintian

almin-stardust:
Signed-off-by: Almin <almin@almin.tf>


JeromeMartinez:
Removing the line is a bit hard, info may be useful.
Please keep the content. Maybe split in 2 lines?

PR #83, opened 2016-09-02
Add PBCore, Mediatrace and image sequence CLI help

kieranjol:
Here's my best guess at what these should look like..


JeromeMartinez:
thanks!

PR #90, opened 2016-10-03
HTML5

retokromer:
- presentation «translated» from XHTML 1.0 to HTML 5
- content identical


JeromeMartinez:
Would not have been what I would have changed first, but why not.
I had great hopes in XHTML, but right it is dead.

retokromer:
Yes, sadly, XHTML is dead. I’m preparing for Albania and came across, therefore…

PR #99, opened 2016-11-27
TIFF extension add

Sami32:
Add TIF and TIFF extensions.

**EDIT**: I could be wrong, but it's look like "tak" extension was mistaped with "tap" in the uninstall part of the _MediaInfo_Extension.nsh_ file, so i corrected it.


Sami32:
Tested:
capture_mediainfo

PR #119, opened 2017-01-23
cUrl

JeromeMartinez:
" --Info_CanHandleUrls" option in the CLI.
returns "1" if URLs can be handled (currently: through libcurl), else "0".


JeromeMartinez:
Depends on https://github.com/MediaArea/MediaInfoLib/pull/374

PR #123, opened 2017-02-05
Add some more extensions supported

Sami32:
WTV are weird files though, i just discoverd that their audio stream was inside the video stream so few tool, like PotPlayer, are able to display the audio stream informations.

The images extensions added are the stereoscopic version.


JeromeMartinez:
WTV is difficult due to lack of specs.

stereoscopic: would you mind to provide some samples of each format?

Sami32:
With pleasure.

Here they are : [3D.zip](https://github.com/MediaArea/MediaInfo/files/753049/3D.zip)
[MPO.zip](https://github.com/MediaArea/MediaInfo/files/753051/MPO.zip)

Sami32:
I don't think it would help you much, but as specifications seem to be very hard to get...
https://en.wikipedia.org/wiki/WTV_(Windows_Recorded_TV_Show)
https://www.itu.int/ITU-T/worksem/h325/200605/presentations/s4p3-levin.pdf
https://msdn.microsoft.com/en-us/library/windows/desktop/dd695085(v=vs.85).aspx

I read that recent MPC-HC release was able of doing that as well (https://github.com/mpc-hc/mpc-hc)
> Ticket #5294, Use internal LAV Splitter for demuxing WTV files

PR #136, opened 2017-03-10
Added VS2017 support and other changes.

kiyolee:
1. Added VS2017 support. Just cloned from MSVC2015.
2. Keep various VS projects/solutions in sync with each other (as much as possible). Fixed file format to what VS would generate.
3. Updated to use wxWidgets 3.1. Including _custom_build project from wxWidgets to remove manual operation before things can build.


JeromeMartinez:
Oops, when I asked about MediaInfo VS2017 update, I actually missed this PR and the new commit reminded me that it was there.
Thank you for the update, I appreciate.

kiyolee:
With the last 2 commits, this branch can now build using wxWidgets 3.1.0 official tarball.
Except that you need to grip wx_config.props from official wxWidgets git source for VS2017 support.
wxWidgets 3.0.2 cannot be used as that does not has support for VS2015 or VS2017.

JeromeMartinez:
Thanks, I'll check and merge next week.

Issue #163, opened 2017-06-18
Am I the only person who finds this thing a pain to work with?


emotionally-exhausted:
Mediainfo is NOT case-insensitive! I type "mediainfo" at the command line and (I'll copy and paste here) it tells me to use "MediaInfo --Help". When I enter this exactly as shown I get "zsh: command not found: MediaInfo
".

I've wasted five minutes trying different variations of upper and lower case and still can't get the help menu to appear. I've used Media info before but abandoned it because it is a major nuisance getting the syntax right. I was hoping somebody would have corrected this by now, but it appears not so I guess I'll have to try and figure out some other way to inject the data into my scripts. Pity, because it is quite the comprehensive program, just not worth the inconsistent and mostly unneeded capitalization.

JeromeMartinez:
> Mediainfo is NOT case-insensitive! 

Actually it depends of the OS, MediaInfo was originally developed on Windows which is case-insensitive.

> I've used Media info before but abandoned it because

It is a bit harsh. Anyway I understand the issue, we just weren't aware of it as there was no report about ot, and we'll fix it (I guess, using the program name instead of hardcoded name).

kieranjol:
Have you added a filename after
`mediainfo`
Eg
`mediainfo /home/emotiontional/exhausted.mov`
?

That prompt that you get when you type lowercase `mediainfo ` is expected.
You haven't provided a file to analyse so what else is mediainfo supposed
to show you?

emotionally-exhausted:
I thought that, like many other programs, it would display a help menu (silly me) that at the very least would show the usage syntax.

JeromeMartinez:
Closed by https://github.com/MediaArea/MediaInfo/pull/164.
Testable with [snapshots](https://mediaarea.net/download/snapshots/binary/).

emotionally-exhausted:
I'm afraid the issue goes far deeper than just the starting of the program.
I've always had problems getting Mediainfo to work properly, which is why I
stopped using it. I'm afraid you'll have to go through the whole program to
make it case-insensitive for Gnu/Linux users.

JeromeMartinez:
> I've always had problems getting Mediainfo to work properly

Thousands of Linux/Mac (case insensitive OS) people already use MediaInfo without issue about any case intensive stuff (and the uppercases in the help part was not a problem), so please stop complaining this way if you want to be listened.

> I'm afraid you'll have to go through the whole program to make it case-insensitive for Gnu/Linux users.

[...]

PR #164, opened 2017-06-20
Help: print real program name

g-maxime:


Source/CLI/CLI_Main.cpp
```diff
@@ -117,6 +117,13 @@ int main(int argc, char* argv_ansi[])
     //Retrieve command line (mainly for Unicode)
     GETCOMMANDLINE();
 
+    //Get real program name, if aviable
+    if(argc && argv_ansi[0] != '\0')
+    {
+        String Program=argv[0];
+        Program_Name(Program);
```
JeromeMartinez: Why using intermediate Program variable?

Source/CLI/Help.cpp
```diff
@@ -11,18 +11,53 @@
...
+            Program = Program.substr(Program.rfind('\\')+1);
+            Program = Program.substr(0, Program.find('.'));
+        #else
+            Program = Program.substr(Program.rfind('/')+1);
+        #endif
+    }
+
+    if(Program.empty())
+    {
+        #if defined(UNICODE)
```
JeromeMartinez: for all "UNICODE" stuff:
there is a "__T()" define, especially for not repeating code (is "L" with Unicode, "" with non Unicode)

Source/CLI/Help.cpp
```diff
@@ -11,18 +11,53 @@
...
+    }
+
+    return Program;
+}
 
 //---------------------------------------------------------------------------
 int Help()
 {
-    TEXTOUT("Usage: \"MediaInfo [-Options...] FileName1 [Filename2...]\"");
+    String Program = Program_Name();
```
JeromeMartinez: No need of intermediate Program variable.

Source/CLI/Help.cpp
```diff
@@ -11,18 +11,53 @@
...
+#include "Common/Core.h"
 #include "Help.h"
 #include "Config.h"
 //---------------------------------------------------------------------------
 
 //***************************************************************************
 //
 //***************************************************************************
+//---------------------------------------------------------------------------
+String Program_Name(const String &Set)
```
JeromeMartinez: Prefer to use a setter and a getter, not both in the same function (if Set is not empty, you don't need the returning String)

g-maxime: With two functions there will be a global variable

Source/CLI/Help.cpp
g-maxime: tanks a lot!

Source/CLI/Help.cpp
JeromeMartinez: Maybe, but I don't see how to do differently.
BTW, tested on Windows and it always shows the default value, looks normal as there is no glbal variable so argv[0] is not "saved". I don't see the idea behind the commit.

JeromeMartinez: lot of `String Program = Program_Name();`, which returns default value, but it should show argv[0].

JeromeMartinez:
Closes https://github.com/MediaArea/MediaInfo/issues/163.
Testable with [snapshots](https://mediaarea.net/download/snapshots/binary/).

PR #168, opened 2017-06-29
cli: help fix

g-maxime:


Source/CLI/CLI_Main.cpp
```diff
@@ -118,7 +118,7 @@ int main(int argc, char* argv_ansi[])
     GETCOMMANDLINE();
 
     //Get real program name, if aviable
-    if(argc && argv_ansi[0] != '\0')
+    if(argc && argv_ansi[0] != "\0")
```
JeromeMartinez: I think I prefer `if(argc && argv_ansi[0] && argv_ansi[0][0])`
(we avoid a pointer in memory with just NULL)

PR #170, opened 2017-07-05
autotools: fix static linking with pkg-config

wiiaboo:
The CMake scripts for libzen and libmediainfo don't use Libs_Static
so LIBS would always be empty. Additionally, the proper way to get
static libcurl libs is to use '--static --libs' when invoking
pkg-config.

To prevent regressions, do both checks for Libs_Static variable
and the new '--static --libs' and use both. It's kind of overlinking,
maybe, but should prevent regressions for now.

### DIFF
--- a/Project/GNU/CLI/configure.ac
+++ b/Project/GNU/CLI/configure.ac
@@ -177,14 +177,15 @@ if test "$with_dll" != "yes"; then
 			LIBS="$LIBS $(libmediainfo-config LIBS)"
 		fi
 	else
-		if pkg-config --exists lib$with_libmediainfo_name; then
+		if pkg-config --exists libmediainfo; then
 			CXXFLAGS="$CXXFLAGS $(pkg-config --cflags libmediainfo)"
 			if test "$enable_staticlibs" = "yes"; then
 				with_mediainfolib="system (static)"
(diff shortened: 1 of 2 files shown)


JeromeMartinez:
Is `$with_libmediainfo_name` an issue?
there is an option https://github.com/MediaArea/MediaInfo/blob/master/Project/GNU/CLI/configure.ac#L55 which would not work anymore.

wiiaboo:
I dunno if it's used for the pkg-config files. I just changed it to be consistent. Since it didn't fail before, it shouldn't fail now, is what I thought.

JeromeMartinez:
If it doesn't hurt, I would like to keep it for the moment, not risking to break something else for another person.

wiiaboo:
Notice how the --cflags check already used the hardcoded name before. Also the GUI configure already used the hardcoded 'libmediainfo' consistently.

JeromeMartinez:
Convinced.
I'll remove the option (it does nothing), and see if someone complain!
(I think I was the only one to use it... 15 years ago, ouch :) )

wiiaboo:
I've just confirmed in libmediainfo and the pkg-config files always use the same libmediainfo.pc name even if you use `--with_libmediainfo_name=whatever`.

PR #213, opened 2018-02-23
Help.cpp - adds OLDXML option

kieranjol:
to prevent annoying questions like this ;) https://github.com/MediaArea/MediaInfoLib/issues/785


JeromeMartinez:
It hurts, but true.
[Some people begins to be interested in getting better documentation](https://mediaarea.net/Vote/Mediainfo-Usage-Documentation-127), I hope to have more time about that in the future, in the meanwhile such PR is appreciated.

PR #218, opened 2018-03-21
typos and fixes

the-j0k3r:
some obvious typos fixes

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,27 +6,25 @@ MediaInfo is a convenient unified display of the most relevant technical and tag
 [![Build status](https://ci.appveyor.com/api/projects/status/msq8b4fuqfy9jmvr/branch/master?svg=true)](https://ci.appveyor.com/project/MediaArea/mediainfo/branch/master)
 
 
-
 # How to build MediaInfo
 
-## Build under Unix (Mac and Linux)
+## Build under MacOS and Linux
 
 First, you must create a directory which will receive the MediaInfo directory, and ZenLib and MediaInfoLib if you decide to compile them yourself.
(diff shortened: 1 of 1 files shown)


JeromeMartinez:
Thanks, appreciated.

Would you mind to change `http://` to `//` (HTTPS version for HTTPS users, HTTP version for HTTP  users) or `https://` (HTTPS version for all) in order to have no use of unencrypted URL for HTTPS users?
Ho, I would actually prefer to use `https://mediaarea.net/MediaInfo/License` instead, I prefer to avoid websites I don't control, when I can.

And as we are changing the Mac part, I suggest to use the preferred Apple form, [macOS](https://en.wikipedia.org/wiki/MacOS);

the-j0k3r:
> Ho, I would actually prefer to use https://mediaarea.net/MediaInfo/License instead, I prefer to avoid websites I don't control, when I can.

I hear you and done. What do you think?

JeromeMartinez:
Perfect, thanks again.
Don't hesitate if you find other typos and fixes.

the-j0k3r:
Depends if you think its worth fixing the odds and ends typos in the `History_CLI.txt` and `History_GUI.txt` documents. at which time it may be worth considering converting those to `markdown` format.

the-j0k3r:
I can surely have a look at the ones you pointed out, if I can be of service and time permitting, you can expect some more contributions.

The README here, generally speaking, could be improved, at the moment though, it satisfies IMHO the basic job it proposes to do, without being hard to understand.

Many hands make light work and many eyes see further.

PR #219, opened 2018-03-21
[en] Obvious typos fix

the-j0k3r:
Hello again, this was a quick cursive look, I notice e.g. Choregrapher -> Choreographer appears in other languages, I can correct those also just will do them in one go instead of one by one.

As for Program settings strings, Ill need to see where strings appear and what type of setting it is. i.e. checkbox/other.

Let me know what you need.

### DIFF
--- a/Source/Resource/Plugin/Language/en.csv
+++ b/Source/Resource/Plugin/Language/en.csv
@@ -516,7 +516,7 @@ Genre_119;Freestyle
 Genre_120;Duet
 Genre_121;Punk Rock
 Genre_122;Drum Solo
-Genre_123;A cappella
+Genre_123;A Cappella
 Genre_124;Euro-House
 Genre_125;Dance Hall
 Genre_126;Goa
--- a/Source/Resource/Plugin/Language/en.csv
+++ b/Source/Resource/Plugin/Language/en.csv
@@ -162,7 +162,7 @@ Album_ReplayGain_Gain;Album replay gain
 Album_ReplayGain_Peak;Album replay gain peak
 Alignment;Alignment
 Alignment_Aligned;Aligned on interleaves
-Alignment_Split;Split accross interleaves
+Alignment_Split;Split across interleaves
 All;All
 AlternateGroup;Alternate group
 Archival_Location;Archival location
@@ -218,7 +218,7 @@ Choose export format;Choose your desired export format
 Choose file(s);Choose the files to open
 Choose filename;Choose your desired filename
 Choose language;Choose your desired language
-Choregrapher;Choregrapher
+Choreographer;Choreographer
 Chroma;Chroma
 ChromaSubsampling;Chroma subsampling
 Close;Close
@@ -516,7 +516,7 @@ Genre_119;Freestyle
 Genre_120;Duet
 Genre_121;Punk Rock
 Genre_122;Drum Solo
-Genre_123;A Cappella
+Genre_123;A cappella
 Genre_124;Euro-House
 Genre_125;Dance Hall
 Genre_126;Goa


JeromeMartinez:
For Choregrapher typo, the string on the left is the "key" in the software, I'll handle fix for other languages when I fix the issue in the software itself.

JeromeMartinez:
Current program settings main window: 

image

thanks again, it is appreciated.

the-j0k3r:
> For Choregrapher typo, the string on the left is the "key" in the software

Yes I figured.



> I'll handle fix for other languages when I fix the issue in the software itself.

No problem. I dont mind, whatever is best for you.

Source/Resource/Plugin/Language/en.csv
```diff
@@ -516,7 +516,7 @@ Genre_119;Freestyle
 Genre_120;Duet
 Genre_121;Punk Rock
 Genre_122;Drum Solo
-Genre_123;A Cappella
+Genre_123;A cappella
```
Sami32: In fact it should be `A capella`
http://id3.org/id3v2.3.0#Appendix_A_-_Genre_List_from_ID3v1

JeromeMartinez: Looks like the typo is on id3.org website.
[A cappella](https://en.wikipedia.org/wiki/A_cappella), with two "p".

the-j0k3r: https://en.wikipedia.org/wiki/A_cappella

Sami32: Yes, it seem that nobody have notified them about that typo yet.
So as we cannot rely on what should be the reference...
https://en.wikipedia.org/wiki/ID3#standard

I do not agree on the capitalized letter as it seem to be the norm:
http://id3.org/id3v2.3.0#Appendix_A_-_Genre_List_from_ID3v1
https://en.wikipedia.org/wiki/ID3#standard

JeromeMartinez: I am in favor to keep the capitalized letter, as it is capitalized  in the "standard"

the-j0k3r: so really what @JeromeMartinez  is saying is that it should remain `A Cappella`? Im consfused now.

What STANDARD are we following after all? This is all I want to know.

JeromeMartinez:
The (de facto) standard is at http://id3.org/id3v2.3.0#Appendix_A_-_Genre_List_from_ID3v1.
But obviously we should not copy the typo (single p). So I think it is better to let as is (removing this as it line from the patch).

the-j0k3r:
> So I think it is better to let as is (removing this as it line from the patch).

Thats what I want to know. Thanks... anyway...

the-j0k3r:
> The (de facto) standard is at http://id3.org/id3v2.3.0#Appendix_A_-_Genre_List_from_ID3v1.
> But obviously we should not copy the typo (single p). So I think it is better to let as is (removing this as it line from the patch).

You mean TWO typos, single p and non Camel Cased... hahaha Standards, all flawed thanks to many people some grammar nazis and some just cant spell anyway, all having fingers in same pie.

Anyway I reverted the line.

Source/Resource/Plugin/Language/en.csv
Sami32: @the-j0k3r Deep a little more and you will see that it is software that support officially the ID3 Tag standard, so they just apply what is in the standard.
http://id3.org/Implementations

Human been human, no blame can be done for following the standard or mistyping as we all do. Just looking at the IEC standards and all their edition addendum make it clear...

The standard ESPECIALLY should be followed.
As this genre in the ID3 standard, and many other ones, come from Winamp, "A Cappella" seem the correct choice.

We are just scratching the surface of what many people knew since very long time.

the-j0k3r: Not digging any deeper, its a futile exercise and besides Mediainfo is actually more correct than that would be standard.

Sami32: Good 👍  it is now ONE more less than at start then ;)

And i'm sure you will correct much more than that later ;)

the-j0k3r: This is the last of them, others already fixed in website and so on.

Sami32: Great, your english skill is much appreciated ;)

the-j0k3r: English, hahahaha

Sami32: Hehe, as you saw we are not english native so your skill is welcome ;)

the-j0k3r: Nor am I... :)

Sami32: Don't tell me that you learned english looking at CNN :)
Frenches and japaneses are known to have poor language skill, hopefully that is not your case ;)
You seem to be in far nordic land ;)

the-j0k3r:
ping, any advances on this?

JeromeMartinez:
Apologizes for the extra long delay in merging your PR.
Changes needed in other files are on the way ([MediaInfoLib](https://github.com/MediaArea/MediaInfoLib/pull/1106), [MediaInfo](https://github.com/MediaArea/MediaInfo/pull/340)).

PR #259, opened 2018-07-02
RPM with version suffix for CentOS/RHEL

g-maxime:


Vascom:
It may be better to use %global onstead of %define.
https://fedoraproject.org/wiki/Packaging:Guidelines?rd=Packaging/Guidelines#.25global_preferred_over_.25define

PR #325, opened 2019-02-27
Android GUI: Reduce apk size

g-maxime:


Source/GUI/Android/app/proguard-rules.pro
```diff
@@ -14,8 +14,13 @@
 
 # Uncomment this to preserve the line number information for
 # debugging stack traces.
-#-keepattributes SourceFile,LineNumberTable
+-keepattributes SourceFile,LineNumberTable
```
JeromeMartinez: May does the opposite (increasing binary size), in a separate PR after dedicated tests.

PR #342, opened 2019-04-12
Cocoa GUI: Update to macOS10.14 SDK with Mojave's dark theme support

g-maxime:


JeromeMartinez:
The .xib changes are due to the SDK upgrade? So we would be no more compatible with older SDKs?

g-maxime:
@JeromeMartinez I had to fix some issues with dark theme (and toolbar) and Xcode cannot convert back to the old xib format. But MyWindow.xib already uses this format.

JeromeMartinez:
> But MyWindow.xib already uses this format.

So compatible with some older SDKs, OK.
We'll see and revert/update if we decide to use and older SDK at some point.

PR #359, opened 2019-06-17
Generate macOS < 10.13 compatible DMG

g-maxime:


JeromeMartinez:
Note: actually it is generating macOS < 10.13 compatible DMG on macOS >= 10.14 (Apple changed their default FS settings and we missed that when we upgraded our build machine), if I well understood our issue.

PR #398, opened 2019-11-28
build: add conditional autoreconf

dabrain34:


Project/GNU/CLI/autogen.sh
```diff
@@ -1,29 +1,10 @@
...
-#automake
-if test -e /usr/bin/automake-1.11 ; then
-  #OpenSolaris: no automake
-  automake-1.11 -a
-elif test -e /usr/bin/automake-1.10 ; then
-  automake-1.10 -a
-else
-  automake -a
+chmod a+x configure
+if [ -z "$NOCONFIGURE" ]; then
```
JeromeMartinez: I am reluctant to have default behavior to launch `./configure`, it is not the previous behavior and `autogen.sh` is planned for generating files for the final user, here we would have to change our build scripts in other places if we don't want to launch unnecessarily `./configure`.

JeromeMartinez:
> But autotools aren't supposed to be run by the final user.

I am fine with that if autotools are installed, especially if it prevents the user to launch our scripts (I understand from the issue ticket that they fail if autotools is installed).

> Extra, this fail on our macOS build environment

Maybe a warning already existing due to the bad `configure.ac`, and failing due to added "--warnings=all"? Maybe good to fix this warning at the same time.

PR #418, opened 2020-02-26
Update Simplified Chinese translation and add string Format_Compression

stevenlele:
- Update Simplified Chinese translation
- Add language string `Format_Compression`


stevenlele:
Please check out my translation toolkit (#419).

JeromeMartinez:
Thank you for the update.
Some questions, having in mind that I have no knowledge about Simplified Chinese:
- for singular (`blabla1` lines), you added `个` so plural and singular become same. Is it intentional and accepted by everyone? I wonder about that as it was not there in previous translations.
- you add the English term e.g. "warp point" because the translated name is rare, do I understand correctly?
- Why are all music genre translations removed?

Remarks:
- "Fast Fusion" is without dash, from https://en.wikipedia.org/wiki/List_of_ID3v1_Genres

stevenlele:
@JeromeMartinez Thank you for your reply.

- There's no difference between singular and plural forms in Chinese. 0 apples, 1 apple and 2 apples are just “0/1/2 个苹果”, where “个” is a general unit for apple (like “a piece of”), and “苹果” is apple. In previous translations, the translators were not aware that these are units and will be appended to a number, so they translated the word itself and missed the “个”. Therefore I wrote some notes in the translation kit.
- For the English terms, you are right. They are usually mentioned in their English form, so either the Chinese translation cannot let user understand the information correctly, or attaching an English abbreviation will let them get the idea more quickly.
- Here are my reasons for removing the genre translations:
  - The music genre translations were taken from the Traditional Chinese translation (Chinese, Taiwan), so the wording is quite different from that in Simplified Chinese.
  - The genre names either do not have a common translation, or they have several translations that might be confusing. [This](https://zh.wikipedia.org/wiki/ID3#.E8.97.9D.E8.A1.93.E9.A1.9E.E5.9E.8B.E5.88.97.E8.A1.A8) is the Wikipedia page for the genres, and you can see many of them do not have a translation, not to mention they are not localized for Simplified Chinese (S. and T. Chinese share the same version of Wikipedia and there is a complex conversion mechanism).
    - In fact, the language names are also taken from the T. Chinese translation. I spent a lot of time searching for the correct translation in S. Chinese to correct them. It was really painful (even with the help of the Wikipedia page - there are a lot of translation variants).
  - AFAIK there isn’t a page written in Simplified Chinese which covers all the translation.
  - Those who care about the genre are usually music professionals. They should have enough knowledge to understand the English names. (similar to the English terms)
  - Tag editors usually have the genre names untranslated, so it is safe to leave the genre names untranslated.

JeromeMartinez:
Thank you for your explanations (and your hard work).
I let the PR open for a few days because I pinged the previous author for comments.
I'll also check the translation kit, it can definitely be useful for others (and me, I was using a old painful code for the merge of files, I may switch to Python... :) ).

stevenlele:
@JeromeMartinez Could you take a look at MediaArea/MediaInfoLib#1217? It should be easy to solve.

JeromeMartinez:
@stevenlele I am updating language files through https://github.com/MediaArea/MediaInfo/pull/427.
There are new fields to translate ;-).
I use this PR also for merging fixes not relating to Simplified Chinese translation, so your PR will be only about Simplified Chinese translation.

I forked your branch, updated it from master + my PR, and squashed your commits, at https://github.com/JeromeMartinez/MediaInfo/commits/update-s-chinese. If fine for you please update your PR with this modified patch + (if you want) update of new language fields.

By the way, consider this [commit](https://github.com/JeromeMartinez/MediaInfo/commit/3937cab52d54bb1816b24c546c8d1f6a4932cc12) in a dedicated branch, it was a proposal of update of Simplified Chinese I forgot to merge, it is conflicting with your PR, maybe you can peek some suggestions from this commit. I won't merge it, just hints for you.

stevenlele:
@JeromeMartinez I was planning to include the translation of the new fields in this PR :joy: Well, I'll open a new PR then.

Thanks for the hints.

JeromeMartinez:
> I was planning to include the translation of the new fields in this PR 😂 Well, I'll open a new PR then.

Oops.
Anyway, the PR was big enough, no big regrets, waiting for the next one. Thank you for you hard work on the Simplified Chinese translation.

Issue #419, opened 2020-02-28
Translation Toolkit


stevenlele:
When I was translating (#418), I did some research and wrote some translation notes.

After that, I wrote a toolkit which can help other translators work more efficiently.

This includes scripts to convert the language file into a spreadsheet file in their target language, together with adding comments and notes.

Please put this toolkit somewhere to help the other translators.

~~[mediainfo_translation_toolkit.zip](https://github.com/MediaArea/MediaInfo/files/4265120/mediainfo_translation_toolkit.zip)~~

~~[mediainfo_translation_toolkit.zip](https://github.com/MediaArea/MediaInfo/files/4271369/mediainfo_translation_toolkit.zip)~~ (Update on Mar 1, 2020: Fix bug)

[mediainfo_translation_toolkit.zip](https://github.com/MediaArea/MediaInfo/files/4271380/mediainfo_translation_toolkit.zip) (Update on Mar 1, 2020: add note to "at ... fps")

Please read `readme.md` for instructions.

JeromeMartinez:
@stevenlele would you mind to open a PR with this files in e.g. `/Contrib/mediainfo_translation_toolkit`?

stevenlele:
@JeromeMartinez OK.

PR #428, opened 2020-03-29
Update Simplified Chinese translation and add translation toolkit

stevenlele:
Translation Toolkit: #419


JeromeMartinez:
Thank you!

PR #440, opened 2020-04-14
Update Simplified Chinese language file

JeromeMartinez:
Update from the translator having started the translation.
Comments from him:

> Punctuation in Chinese is divided into half-width and full-width.
[,.!?""]half-width　Used for English
【，。！？“”】full-width　Used for Chinese characters
Add or not add spaces between English letters and Chinese characters.
zh-CN
See personal habits. There are no mandatory standards.
The translated text has been unified as: Chinese punctuation, with spaces between  English letters and Chinese characters.

The changes I have made:
- name and email of translator are the ones from both current translators.
- It seems there is a debate about left column width in text view (30 vs 40); ad zh-CN seems a lot shorter than European languages, I currently keep 30.

@stevenlele please review and comment.

Please note that I have absolutely no idea about what are the changes and what is the "best", I am just trying to have the best translation file for my zh-CN users.


stevenlele:
@JeromeMartinez 

Apparently his changes are based on the previous incomplete translation (his own version) and he only added the missing entries. So the result is that he reverted most changes that I made. I strongly disagree with these changes. After going through them, I don't think I see any good edit.

As you see, I spent a lot of time doing research into every entry, so I fully understand the meaning of every entry. I searched on Internet and digged into the source code. The translation toolkit is the proof. Reducing the column width is also a proof. Apparently he did no research, nor did he check my toolkit, so his changes brought a lot of mistakes back. Not to mention I have decent knowledge about media files - I know video processing (e.g. VapourSynth), encoding (e.g. x264, x265), container formats, muxing and demuxing (e.g. mp4 and mkv) etc. You see, I included doom9 links in the toolkit. I deal with media files a lot, so I know what people use MediaInfo for, and what terms are appropriate.

Moreover, I myself have much exprience in translating. I helped translate other computer softwares and I know how to avoid bad translation. No offense here, but I think my English expression is better and more fluent than his.

Regarding the punctuation styles, I go with half-width ones because they make the information look shorter. In my opinion, easy to read is more important than style preference. But after all, the punctuation he used himself was inconsistent. He used half-width question marks and parentheses. I guess it's because if he used full-width ones, the strings would be too long and too "spacious". I still recommend half-width ones.

I know that detailed explanation about each entry is needed to convince you, but according to my experience, it would be too time-consuming, not to mention the time that I already spent on the translation work. **My suggestion is to reject this change.** I believe he had better compare my translation, his translation and the original English text, check my translation toolkit, and provide reason for each change.

Thank you.

JeromeMartinez:
I currently don't have strong opinion, more on discarding this change except one thing I forgot to mention from feedback:

> Music genre I still want to use Chinese + original English.

It may be good to have this system of Chinese + original English, I don't see how it could hurt a lot.

stevenlele:
My opinion from https://github.com/MediaArea/MediaInfo/pull/418#issuecomment-593798668:
> Here are my reasons for removing the genre translations:
>   
>  * **The music genre translations were taken from the Traditional Chinese translation (Chinese, Taiwan)**, so the wording is quite different from that in Simplified Chinese.
>  * The genre names either do not have a common translation, or they have several translations that might be confusing. [This](https://zh.wikipedia.org/wiki/ID3#.E8.97.9D.E8.A1.93.E9.A1.9E.E5.9E.8B.E5.88.97.E8.A1.A8) is the Wikipedia page for the genres, and you can see many of them do not have a translation, not to mention they are not localized for Simplified Chinese (S. and T. Chinese share the same version of Wikipedia and there is a complex conversion mechanism).
>     
>    * In fact, the language names are also taken from the T. Chinese translation. I spent a lot of time searching for the correct translation in S. Chinese to correct them. It was really painful (even with the help of the Wikipedia page - there are a lot of translation variants).
>  * AFAIK there isn’t a page written in Simplified Chinese which covers all the translation.
>  * Those who care about the genre are usually music professionals. They should have enough knowledge to understand the English names. (similar to the English terms)
>  * Tag editors usually have the genre names untranslated, so it is safe to leave the genre names untranslated.

I'm not against Chinese + English, but he had better do thorough research and figure out the correct translation in Simplified Chinese. Using the current "converted" translation from Traditional Chinese is similar to using Brazilian Spanish for Castilian Spanish, etc. It's just not right.

PR #515, opened 2021-01-06
configure.ac: Use PKG_CONFIG variable before using pkg-config

1480c1:
This can solve a very specific issue where pkg-config might be pkgconf and a specific flag might be needed for the output to match pkg-config


JeromeMartinez:
If MediaInfo has such issue, hasn't MediaInfoLib the same issue?

1480c1:
It seems MediaInfoLib also has the issue, but since I use cmake with it, I have not encountered it yet

g-maxime (approved):

1480c1:
Thanks!

JeromeMartinez:
@1480c1 appreciated that it is more versatile, don't hesitate to send patches for improving our build scripts.

PR #578, opened 2021-11-18
Use system icons instead of built-in

buckmelanoma:
This PR replaces MediaInfo's built-in icons with icons from the local system theme in the Qt version.

Screenshot_20211118_144047


JeromeMartinez:
We don't use much this version upstream, thanks for the suggestion, merged.

PR #579, opened 2021-11-28
Add removable-media plug

kaytat:
https://snapcraft.io/docs/removable-media-interface

This interface will allow mediainfo to access mounted media. This plug is not auto-enabled and requires and explicit command to activate.


kaytat:
I did not try building this locally, however.

g-maxime (approved):

PR #595, opened 2022-02-21
remebers view on Linux GUI

BlackenedSky84:
Saves the current view on exiting the restores view on load in WxWidgets version.
can be disabled in config file.


JeromeMartinez (approved):
Thank you for your PR.
We definitely need to improve the way to get the list of views, but this is out of the scope of such PR and we'll adapt it when we have something that can be shareable across all UIs.

PR #642, opened 2022-09-15
Windows GUI: Switch to https for MediaArea URLs

g-maxime:


JeromeMartinez (approved):
Changelog check uses now HTTPS, with TLS 1.2 instead of TLS 1.3 available but fair enough for the moment.

PR #652, opened 2022-10-05
Zh-cn translation update

nkh0472:
Update Simplified Chinese language file


JeromeMartinez:
@stevenlele please review and comment about translation of new fields.

PR #653, opened 2022-10-18
Update Simplified Chinese translation

stevenlele:
Follow-up of #652.

Fixed and added translation. Also added some notes to the translation toolkit.

Some terms are kept in English because they are either hard to translate or better understood by professionals:
- Dolby Atmos: Downmix and Trim ([ref](http://web.archive.org/web/20220128060439/https://learning.dolby.com/hc/en-us/articles/360054531672-Module-7-4-Renderer-Trim-and-Downmix-Controls-))
- Dolby Audio: Balance
- AC4: gain (Four backs to two backs, Screen to center, etc)


JeromeMartinez:
Greatly appreciated!

PR #658, opened 2022-11-10
Fix MSVC2022 project Includes

RytoEX:
The MSVC2022 project Includes point to the MSVC2019 projects for MediaInfoLib and ZenLib. Point to the MSVC2022 projects instead to prevent build failures on systems with only the v143 toolset.


JeromeMartinez:
Oops, thanks!

PR #683, opened 2023-03-14
Qt GUI: HighDPI support

Kolcha:
avoid scaling issues and blurred icons in menus on HighDPI screens
screenshots from KDE, left - no changes (master as is), right - with this fix (noticeable change for round icons)

### DIFF
--- a/Source/GUI/Qt/main.cpp
+++ b/Source/GUI/Qt/main.cpp
@@ -27,6 +27,10 @@ int main(int argc, char *argv[])
     QCoreApplication::setApplicationName("mediainfo-gui");
     QCoreApplication::setApplicationVersion(QString().fromStdString(Ztring(MediaInfo_Version_GUI).To_UTF8()));
 
+#if QT_VERSION >= QT_VERSION_CHECK(5, 6, 0) && QT_VERSION < QT_VERSION_CHECK(6, 0, 0)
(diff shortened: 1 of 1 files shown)


Source/GUI/Qt/main.cpp
```diff
@@ -27,6 +27,10 @@ int main(int argc, char *argv[])
     QCoreApplication::setApplicationName("mediainfo-gui");
     QCoreApplication::setApplicationVersion(QString().fromStdString(Ztring(MediaInfo_Version_GUI).To_UTF8()));
 
+#if QT_VERSION >= QT_VERSION_CHECK(5, 6, 0) && QT_VERSION < QT_VERSION_CHECK(6, 0, 0)
```
JeromeMartinez, line 30: They removed it in Qt6 ([v6 doc](https://doc.qt.io/qt-6/qt.html#ApplicationAttribute-enum) vs [v5 doc](https://doc.qt.io/qt-5/qt.html#ApplicationAttribute-enum)) so source code not compatible???

Is there a way to do same with Qt6 or is it automatic in Qt6?

Kolcha: yes, you are right, with Qt6 these options are by default. they are needed only in case of Qt5, so that's why I placed them under version check macro

JeromeMartinez:
Thank you for the update.

PR #684, opened 2023-03-14
Qt GUI: fixed QT_VERSION macro usage

Kolcha:
- [values for `QT_VERSION`](https://doc.qt.io/archives/qt-4.8/qtglobal.html#QT_VERSION) were incorrect (expected Qt 64.x.y instead of Qt 4.x.y)
- drop invalid code under such incorrect version comparison
- drop [deprecated function](https://doc.qt.io/qt-5/qlayout-obsolete.html#setMargin) that was added since Qt 5, but [deprecated in 5.13](https://code.qt.io/cgit/qt/qtbase.git/tree/src/widgets/kernel/qlayout.h?h=v5.15.8-lts-lgpl#n86)
  correct equivalent already existed in the code, just was not called due to incorrect Qt version check

### DIFF
--- a/Source/GUI/Qt/configtreetext.cpp
+++ b/Source/GUI/Qt/configtreetext.cpp
@@ -134,7 +134,7 @@ void ConfigTreeText::addField(int i, QString f) {
 }
 
 void ConfigTreeText::removeField(int i, QString f) {
-#if QT_VERSION >= 0x404000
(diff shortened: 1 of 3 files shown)


JeromeMartinez:
Thanks!
There are definitely lot to do on the Qt related code :(.

PR #689, opened 2023-03-18
Qt GUI: get rid of obsolete QRegExp usage

Kolcha:
replaced QRegExp with QRegularExpression, it enables Qt 6 build (no any further changes are required), but breaks compatibility with Qt 4 (what is not big deal as for me)

in these particular cases it was enough jsut to replace, no changes to patterns are required


JeromeMartinez:
Thank you for the update.
Qt4 support is not expected to be still possible :-p.

PR #715, opened 2023-04-21
Qt GUI: About dialog fixes

Kolcha:
- simplified layout without visual changes
- better resize behavior - keep "about text" aligned to top
- use SVG icon for "info" icon (HighDPI support)
- "OK" button is default button
- get app version "in place" instead of passing it
- use format string for "about text" instead of "keyword + replace"

Screenshot_20230420_082829Screenshot_20230421_022932


JeromeMartinez:
Thanks.

PR #734, opened 2023-06-21
Update eu.csv

alexgabi:
Please update basque language (eu)
Thanks

### DIFF
--- a/Source/Resource/Plugin/Language/eu.csv
+++ b/Source/Resource/Plugin/Language/eu.csv
@@ -1,6 +1,6 @@
   Language_ISO639;eu
-  Author_Email;azpidatziak@gmail.com
-  Author_Name;Xabier Aramendi
+  Author_Email;alexgabi@disroot.org
+  Author_Name;Alexander Gabilondo
   Author_OldNames;
   Language_Name;Euskara
   Config_Text_ColumnSize;40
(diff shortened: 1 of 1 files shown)


Source/Resource/Plugin/Language/eu.csv
```diff
@@ -1,6 +1,6 @@
   Language_ISO639;eu
-  Author_Email;azpidatziak@gmail.com
-  Author_Name;Xabier Aramendi
```
JeromeMartinez, line 3: Put the older name in `Author_OldNames` field.

Source/Resource/Plugin/Language/eu.csv
```diff
@@ -11,46 +11,46 @@
...
- file1; agiri
- file2; agiri
- file3; agiri
- fps1; fs-ko
- fps2; fs-ko
- fps3; fs-ko
- frame1; frame
- frame2; frame
- frame3; frame
+ bit1; bit1
```
JeromeMartinez, line 14: Seems a regression, I doubt that you want "1" and so on displayed (several times, bit, Byte, channel...).
Remove the numbers.

Source/Resource/Plugin/Language/eu.csv
```diff
@@ -1162,12 +1161,12 @@ SamplesPerFrame;
 SamplingCount;Lagin zenbatekoa
 SamplingRate;Laginketa neurria
 Save;Gorde
-ScanOrder;Mihaketa hurrenkera
```
JeromeMartinez, line 1165: Current translation is so wrong? (several times, when you remove completely a translation without a replacement)

JeromeMartinez (changes requested):
Thank you for the update proposal.
Below are some comments.

Source/Resource/Plugin/Language/eu.csv
alexgabi: I forgot it. I will do it.

Source/Resource/Plugin/Language/eu.csv
alexgabi: Some of the more technical terms were not well translated and, in the case of those that are not usually used by non-expert users, the original term is usually maintained.

JeromeMartinez:
Superseeded by https://github.com/MediaArea/MediaInfo/pull/928 (and same on me to have let this one open forever).

PR #777, opened 2023-11-19
Add ffmpeg plugin

g-maxime:


kieranjol:
What does this do?

JeromeMartinez:
> What does this do?

My part of the work is late, at the end it is for showing active display (width/height). like [Active Format Description](https://en.wikipedia.org/wiki/Active_Format_Description) but without this metadata, in the MediaInfo report in addition to the coded width/height, without the user needing to have 2 tools, one being a command line.

Issue #672, opened 2023-01-27
Mediainfo Dark Mode 


userzzzq:
support win10/win11 dark mode

Thermalzombie:
Would like this to please.

Thermalzombie:
Could you not at lease add an option for the text view to set black background white font.  Thanks?

cjee21:
Looks like this is possible to implement. Just need to figure out how to detect current Windows mode and switch at runtime.

![Screenshot 2024-05-19 161718](https://github.com/MediaArea/MediaInfo/assets/77721854/17742d48-5276-4a38-9625-3ccf24de32de)
![Screenshot 2024-05-19 162138](https://github.com/MediaArea/MediaInfo/assets/77721854/8cd54d4f-c931-4d1d-9330-f13f1e74b4e6)

cjee21:
I have implemented a very basic dark mode that can be toggled in the Options menu with the following limitations:
- It does not follow Windows current mode.
  - This requires checking the current Windows mode on application startup as well as detecting changes to Windows mode while the application is running which I am not able to implement.
- It does not remember the user preferred mode and will always default to light mode.
- The MediaInfo logo in the About window is not visible in dark mode because it is black.
- Some pages might still be displayed in light mode.

Do you want me to make a pull request for this?

![Screenshot 2024-05-19 175555](https://github.com/MediaArea/MediaInfo/assets/77721854/a9c9acb4-0be8-4351-9fc7-a72705ab509d)

PR #836, opened 2024-05-19
Basic dark mode for Windows GUI

cjee21:
A very basic dark mode for Windows GUI that can be toggled in the Options menu.

It has the following limitations:
- It does not follow Windows current mode.
  - This requires checking the current Windows mode on application startup as well as detecting changes to Windows mode while the application is running which I am not able to implement.
- It does not remember the user preferred mode and will always default to light mode.
- The MediaInfo logo in the About window is not visible in dark mode because it is black.
- Some pages might still be displayed in light mode.


JeromeMartinez:
`Windows11_Modern_Dark.vsf` is not on our build machine (also not on my home machine), maybe installed only on Win11 machines? I don't want Win11 to be mandatory for building the tool.
Please use e.g. `Windows10Dark.vsf`.

cjee21:
> `Windows11_Modern_Dark.vsf` is not on our build machine (also not on my home machine), maybe installed only on Win11 machines? I don't want Win11 to be mandatory for building the tool. Please use e.g. `Windows10Dark.vsf`.

I got that from the Getit Package Manager as mentioned in first post.
Screenshot 2024-05-21 215743

If it is not possible to add this then I can switch to Windows10Dark.vsf. However, Windows10Dark.vsf has two issues:
- It is totally black so MediaInfo logo is not visible in ABout window and maybe also the title bar.
- It does not have rounded corners on the windows when run on Windows 11.

JeromeMartinez:
> If it is not possible to add this then I can switch to Windows10Dark.vsf.

Oops, I missed that, thanks for the reminder.

>  Windows10Dark.vsf has two issues:
    It is totally black so MediaInfo logo is not visible in About window and maybe also the title bar.

Annoying.
Let's try with `Windows11_Modern_Dark.vsf`, @g-maxime please install it on the build machine.

JeromeMartinez:
[this build](https://mediaarea.net/download/snapshots/binary/mediainfo-gui/20240521-5/MediaInfo_GUI_24.04.20240521_Windows.exe) has this patch.
@cjee21 please rebase from our main branch.

cjee21:
I'm working on getting it to follow Windows dark mode settings. Will re-base and add one more commit if successful.

JeromeMartinez:
> I'm working on getting it to follow Windows dark mode settings.

Appreciated!

JeromeMartinez:
> I'm working on getting it to follow Windows dark mode settings.

Please keep the manual change option, at least in preferences if not in the menu directly.

JeromeMartinez:
> No idea how to pass the current mode over to the other file. I'll get it pushed here first and maybe you can have a look.

It is more classic stuff on our side, something we could manage.

cjee21:
> It is more classic stuff on our side, something we could manage.

The problem is that the code to set dark mode on application start is in MediaInfo_GUI.cpp while the manual toggle is in GUI_Main.cpp. Can't find a way to get current mode. It's my first time using C++Builder and doing VCL. I tried putting the code to set dark mode on application start in GUI_Main.cpp at first but can't find a place for it that works properly.

JeromeMartinez:
> It's my first time using C++Builder and doing VCL

Kudos for having taken the time to learn it and check doc about the different features.

> I tried putting the code to set dark mode on application start in GUI_Main.cpp at first but can't find a place for it that works properly.

I didn't touch this code in the last several years, I don't remind but I'll give it a look. Maybe not this week, however.

cjee21:
> Kudos for having taken the time to learn it and check doc about the different features.

Glad to be able to contribute to MediaInfo while learning useful things along the way and getting more familiar with Git/GitHub.

> I didn't touch this code in the last several years, I don't remind but I'll give it a look. Maybe not this week, however.

Take your time. I am not in a hurry to get this merged. I don't really need dark mode, just helping implement it since it seems doable for me.

cjee21:
Good news! I managed to solve the issues with Windows dark mode integration. I will update this pull request once I finish integrating it into MediaInfo.

cjee21:
Basic dark mode implementation is completed. I have updated the first post to reflect the current state of this PR so please take a look.

One possible improvement is to enable the user to choose if they want to follow Windows dark mode setting or set MediaInfo to always be in light/dark mode. I will not be implementing this as I am not familiar with how MediaInfo handles preferences.

JeromeMartinez:
> One possible improvement is to enable the user to choose if they want to follow Windows dark mode setting or set MediaInfo to always be in light/dark mode. I will not be implementing this as I am not familiar with how MediaInfo handles preferences.

Fair enough, we'll manage that (should be for other things too, we'll implement it when we have time, except if you are motivated for that ;-) ).

> I also do not know if this is good enough to consider issue https://github.com/MediaArea/MediaInfo/issues/672 resolved so I did not mark this as closing that issue.

It is, flagged.

[build](https://mediaarea.net/download/snapshots/binary/mediainfo-gui/20240522-3/MediaInfo_GUI_24.04.20240522_Windows.exe).

> The file open dialog is not high DPI.

Why? Not handled by a component?
And it is not the only issue, the dark mode is buggy there:
image
But I guess this is not in your control. Or maybe you can manage to have another open windows (maybe hight DPI aware)?
Not blocking.

But (see next post)...

JeromeMartinez (changes requested):
For whatever reason (I don't see what when reading the code change), drag and drop is forbidden after this patch, and it is a big blocker.

cjee21:
I guess both blurry open window in high DPI mode and no drag and drop in dark mode is caused by the VCL style. Drag/drop in light mode still works.

JeromeMartinez:
> Drag/drop in light mode still works.

Not on my side, so it seems to be due to a part of the code which breaks that, there is a need to add code step by step and see what breaks the drag and drop.

> 🤔 Is this the reason/solution? https://stackoverflow.com/a/10337744

Please try :)

> I guess both blurry open window in high DPI mode [...] is caused by the VCL style. 

I think that it is due to an open box not supported by the style manager, but it is maybe because this is not a "good" open box. I see that C++ Builder itself has support of dark mode and add a good support of dark mode with the open box, which is different. I would be fine to switch to this open box.

Would you mind to see how to have the other open box and change to it?

PS: this is the exact reason we were reluctant to implement it on our side with impact on our sponsors requests, "just some lines of code, quick" is the expected duration then reality shows up 😄.

cjee21:
Drag and drop should work in all situations now :)

cjee21:
That's one more limitation gone. First post updated. The open file dialog is now available in both dark mode and high DPI but the open folder dialog is not available in dark mode. This is now Windows' fault for not yet implementing dark mode for some dialogs. 😆

cjee21:
By the way, should I make another PR for changing the font of 'Text' view to a newer, nicer font? I have already implemented it. It is because I noticed the current font does not look good and scale well on high DPI settings.

JeromeMartinez:
> By the way, should I make another PR for changing the font of 'Text' view to a newer, nicer font?

I am fine with that, if it stays monospace.
Generally speaking, we are interested in lot of improvement for the Windows GUI, we didn't touch it for years due to priority elsewhere (more on the core of MediaInfo = format detection/analysis), so don't hesitate to provide hints about what you would like to implement and we say if we could accept it or not.

JeromeMartinez (approved):
That is great, thank you.

> [...] the open folder dialog is not available in dark mode. This is now Windows' fault for not yet implementing dark mode for some dialogs.

Also on Windows 11? Crazy... It is not the most difficult to update, but I guess that it is not used enough for MS to care about that...

cjee21:
> (just to be transparent: we are thinking to switch to Qt on Windows in order to have a fully open source workflow + removing one UI toolkit because Qt is already used on Linux and is more cross platform, but well, we are thinking about that for several years now, so it won't be immediate, and we are not yet sure Qt will fit at 100% on Windows)

Noted.

> Also on Windows 11? Crazy... It is not the most difficult to update, but I guess that it is not used enough for MS to care about that...

Even this file properties window is not implemented....
Screenshot 2024-05-22 223442
You can see the background is file explorer which shows dark mode is enabled.

JeromeMartinez:
> open folder dialog is not available in dark mode.

while working with VirtualBox, I see that there is another "open folder" windows! And in dark mode!
("Dossier" means "folder")

image

It seems to be a system one (2 other windows in the screenshots seems to be styled by VirtualBox itself internally, the "open folder" window seems to be from the system with a different style), maybe there is something callable directly in C++ Builder or there is a system call to find...

> Even this file properties window is not implemented....

And this one is used by a lot of people... Not serious like Apple is...

JeromeMartinez:
> It is only tested on Windows 11 and it is not known how it looks on other Windows versions

Tested on Win10 and is totally fine.

On Win7 it is mixed:
image
But expected (software style is by C++ Builder and integrated in their libraries, as for Qt, and the system windows are not aware of dark mode).

cjee21:
@JeromeMartinez Just to let you know, I just discovered there are still some bugs with this dark mode. I will not be fixing them as I have no idea what is the cause.
1. Every time dark mode is toggled, the currently open file will be re-added to the list of open files.
Screenshot 2024-05-23 143101
3. If dark mode is toggled while in another view such as text view, an error may pop-up and the view will switch back to the default view.
Screenshot 2024-05-23 142745

JeromeMartinez:
> I just discovered there are still some bugs with this dark mode. 

Thanks for letting us know, we'll fix that before the release.

Thermalzombie:
Hi thanks for the efforts to bring dark mode.  Just wanted to no what version I need of mediainfo for it to be available.

JeromeMartinez:
>  Just wanted to no what version I need of mediainfo for it to be available.

The next public version or the build listed in https://github.com/MediaArea/MediaInfo/pull/836#issuecomment-2124663164.

cjee21:
> > Just wanted to no what version I need of mediainfo for it to be available.
> 
> The next public version or the build listed in [#836 (comment)](https://github.com/MediaArea/MediaInfo/pull/836#issuecomment-2124663164).

But the build in that comment doesn't include the drag & drop or dialog fix right?

JeromeMartinez:
> But the build in that comment doesn't include the drag & drop or dialog fix right?

Right.
[latest build](https://mediaarea.net/download/snapshots/binary/mediainfo-gui/20240522-6/MediaInfo_GUI_24.04.20240522_Windows.exe).

cjee21:
For opening folders, maybe can consider using this other 'Select Folder' dialog instead. This screenshot is from Mp3tag.
Some advantages include dark mode support, ability to copy paste directory path and view directory history in the address bar, ability to select whether to include subdirectories, as well as possibility of having thumbnail view of folders and viewing folder contents.

Screenshot 2024-05-23 181345

I looked at MediaInfo code and see that it appears to use a complex custom implementation from ZenLib for handling the folder opening so it is not something that I will touch.

JeromeMartinez:
> I looked at MediaInfo code and see that it appears to use a complex custom implementation from ZenLib for handling the folder opening so it is not something that I will touch.

ZenLib should not be an issue there, just dealing with provided path from the system dialog, please provide a PR with a PoC and I check if there are issues.

cjee21:
> ZenLib should not be an issue there, just dealing with provided path from the system dialog, please provide a PR with a PoC and I check if there are issues.

Took some time to figure out but turns out to be very simple. 😮 😅
Making PR now...

PR #839, opened 2024-05-22
Windows GUI: Change monospaced font

cjee21:
This patch makes MediaInfo use Cascadia Mono font for views that use monospaced font if Cascadia Mono is available on the system. Cascadia Mono is the default font for Windows Terminal. Most updated Windows systems with Windows Terminal installed will have this font.

Reason for this change is that the current font does not scale properly. Notice that it does not scale as linearly as Cascadia Mono. Cascadia Mono also has smoother edges and is easier to read especially on high DPI scaling.

Comparison @ 100%
Screenshot 2024-05-22 201854

Comparison @ 150%
Screenshot 2024-05-22 201854

Comparison @ 200%
Screenshot 2024-05-22 202101


JeromeMartinez:
> Let me know if it is desirable to change the font for other views as well.

Yes, I prefer coherence so same font everywhere.

JeromeMartinez (changes requested):
> Let me know if it is desirable to change the font for other views as well.

Yes, I prefer coherence so same font everywhere.

and rebase please.

PR #840, opened 2024-05-22
Windows GUI: Improve window size for high DPI

cjee21:
Take DPI scaling into account when setting window size. This should prevent cases where one ends up with a very small window filled with large icons and text when using a very high DPI display with very high DPI scaling.

Before and after this patch on a QHD display with 150% scaling for example
Screenshot 2024-05-23 005650
The difference will be more with higher scaling.


JeromeMartinez (changes requested):
Keep the current code, then your `DPIScale=` line then a new line with ` Width*=DPIScale;`
Ne need to repeat code with the multiplication.

cjee21:
Tested on QHD display at 100%, 125%, 150%, 175%, and 200%.

JeromeMartinez (approved):

PR #842, opened 2024-05-23
Windows GUI: Dialog fixes

cjee21:
Two commits:

1. Noticed that there is code to handle opening multiple files simultaneously but the Open File dialog doesn't allow selecting multiple files. It is now enabled and multiple files can be selected and opened simultaneously.

2. Minor fix to the positioning of the new Open Folder dialog. It now consistently opens at the same position near the corner of MediaInfo's main window.


JeromeMartinez (approved):

cjee21:
Gentle reminder that this is not merged yet in case it is forgotten. ☺️

PR #843, opened 2024-05-24
Linux GUI: Change to transparent icons

### LINKED ISSUE #350: replace old icons with solid background
this is what it looks like on linux with dark desktop. There are already better icons in mediainfo git, but they seem to be used elsewhere. I suggest replacing these dated ones with the newer ones that have transparent background

cjee21:
Replace the three icons in the toolbar with newly created versions that have transparent backgrounds.

Resolves #350 

Screenshot from 2024-05-24 17-04-51


cjee21:
@JeromeMartinez seems nothing else is using the xpm icon files. If it is not necessary to retain the old icon files, this commit can be simplified to only replace the icon files without requiring any code changes.

cjee21:
> Yes, please.

Done. ☺️

JeromeMartinez (approved):

Issue #741, opened 2023-07-06
Save export with same filename and location as original.


Snake883:
Save the export file name with the same name as the file.
Save the export in the same folder location as the original file.
Only the extension would be different (.txt, .nfo, etc).

JeromeMartinez:
We do that with e.g. [BWF MetaEdit](https://mediaarea.net/BWFMetaEdit), it makes sense to do that with MediaInfo too.
When we have a bit of free time...

cjee21:
Made a PR for this. It'll reduce the amount of clicks and typing required for my use case of export.

PR #847, opened 2024-05-28
Windows GUI: Use file folder and name as export default for single file

cjee21:
When only a single file is open in MediaInfo, use that file's name and folder as the default when exporting instead of MediaInfo's AppData directory.

Not sure if can consider issue #741 as resolved since this PR is only for when there is a single file open.


JeromeMartinez (changes requested):
there is a risk to overwrite the original file (there is now a warning if the file exist, but default to an existing file is a source of problems as people use to click on every window box).
Please append ".MediaInfo.xml" (if XML output, else ".txt" etc).

cjee21:
Made the change. Now it will retain the file extension and add .mediainfo before the exported file extension. So even if user has files with same name but one is .mp4 and another .mkv for example, there is also no risk of overwrite.

JeromeMartinez:
> Now: my test file.mp4.mediainfo.txt

Oops, I missed that the extension is already there.
Please "mediainfo" --> "MediaInfo" (I am a uppercase letters addict).

JeromeMartinez (approved):

PR #848, opened 2024-05-29
Windows GUI: Sheet view improvements

### LINKED ISSUE #500: "Sheet" UI elements should allow resizing and moving.
New user again.  I was surprised by the UI I see when picking "Sheet".  I have a large empty, white space where the filename is listed at the top (is a little more than half the window).  Then 5 lines (only 4 with values), then a larger text area at the bottom about the same size as the 5 lines.  That area at the bottom has a scroll bar, so I assumed I could drag the top line of that pane and it would get bigger.  Using up that wasted white space at the top.  But nope...I must scroll the 6 or so lines of text I can see if I want to know more of what the bottom pane has.

Seems like a waste of screen real estate currently, but maybe I don't know the purpose of that page?

And why can't I change the size of these elements?  I recognize the 5 lines with "web" buttons to the right probably won't change size, but they should move.

cjee21:
Change sheet view default split ratio and make it resizable. Now it shows more of the bottom contents by default and also after maximizing window.

Also change default monospaced font to match other views. Missed this page when changing the default monospaced font the other day.

Resolves #500

### DIFF
--- a/Source/GUI/VCL/GUI_Main.cpp
+++ b/Source/GUI/VCL/GUI_Main.cpp
@@ -216,6 +216,7 @@ __fastcall TMainF::TMainF(TComponent* Owner)
         monoFont->Size = 10;
         Page_Text_Text->Font = monoFont;
         Page_Custom_Text->Font = monoFont;
+        Page_Sheet_Text->Font = monoFont;
     }
 }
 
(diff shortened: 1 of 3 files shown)


cjee21:
Amended to fix the text view string bug that turns out to also affect this view.

JeromeMartinez (approved):

PR #851, opened 2024-05-30
Windows GUI dark mode: avoid an unexpected popup

JeromeMartinez:
See https://github.com/MediaArea/MediaInfo/pull/836#issuecomment-2126335967.

### DIFF
--- a/Source/GUI/VCL/GUI_Main.cpp
+++ b/Source/GUI/VCL/GUI_Main.cpp
@@ -2047,6 +2047,7 @@ void __fastcall TMainF::Footer_ButtonClick(TObject *Sender)
 //---------------------------------------------------------------------------
 void __fastcall TMainF::M_Options_DarkmodeClick(TObject* Sender)
 {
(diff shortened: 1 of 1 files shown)


cjee21:
This does prevent the popup although it still changes views. All the issues seem to be with VCL's implementation of styles which re-spawns the entire window. Unfortunately wihout using styles it seems difficult to make dark mode in VCL. Non-VCL applications can change dark mode without closing and reopening window and they also have native translucent Windows titlebar in dark mode.

The change in this PR needs to be applied to ApplicationEvents1OnSettingChange too to prevent the same popup for when the mode is changed using Windows settings while MediaInfo is running.

cjee21:
Bug number 1 in https://github.com/MediaArea/MediaInfo/pull/836#issuecomment-2126335967 seems to only happen for files which are open via File Explorer context menu.

cjee21:
I found a possible solution for the files duplicating on dark mode change bug mentioned above. I will make a PR later.

PR #852, opened 2024-05-31
Windows GUI dark mode: avoid an unexpected popup 2

JeromeMartinez:
Due to https://github.com/MediaArea/MediaInfo/pull/851#issuecomment-2141248535

### DIFF
--- a/Source/GUI/VCL/GUI_Main.cpp
+++ b/Source/GUI/VCL/GUI_Main.cpp
@@ -2062,7 +2062,8 @@ void __fastcall TMainF::ApplicationEvents1OnSettingChange(
     TObject* Sender, int Flag, const UnicodeString Section, int &Result)
 {
     if (Section == "ImmersiveColorSet") {
-        if (WindowsDarkModeEnabled()) {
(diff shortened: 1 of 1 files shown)


cjee21:
@JeromeMartinez turns out the new PR that I'm going to make (soon)addresses the root cause of all the dark mode switching bugs. This and #851 is no longer needed with the new PR.

JeromeMartinez:
> This and https://github.com/MediaArea/MediaInfo/pull/851 is no longer needed with the new PR.

We do a release today with them as a workaround, we don't want to block the release but it will be in the next one.

> turns out the new PR that I'm going to make (soon) addresses the root cause of all the dark mode switching bugs. 

It would be great! I thought it was a limitation of C++ Builder but maybe it is not :).
Thank ou for taking care of that.

cjee21:
I can make it within an hour if you want to wait.

JeromeMartinez:
> I can make it within an hour if you want to wait.

OK, we wait a bit.

cjee21:
Done. I make a PR for this main issue first and will look at whether it's possible for HTML view to have dark mode for future release.

JeromeMartinez:
> [...] and will look at whether it's possible for HTML view to have dark mode for future release.

Your help is appreciated!

PR #853, opened 2024-05-31
Windows GUI: Prevent repeated execution of certain code

cjee21:
Ensure code that is only necessary to execute once on application start-up is not repeated every time FormShow is called such as when switching dark mode.

Resolves bugs mentioned in https://github.com/MediaArea/MediaInfo/pull/836#issuecomment-2126335967

Also makes dark mode switching faster.

PR #851 and #852 no longer needed. Now the view will not change when switching dark mode.

### DIFF
--- a/Source/GUI/VCL/GUI_Main.cpp
+++ b/Source/GUI/VCL/GUI_Main.cpp
@@ -332,31 +332,35 @@ void __fastcall TMainF::GUI_Configure()
 void __fastcall TMainF::FormShow(TObject *Sender)
 {
     //Configuration of MediaInfoLib
-    if (I==NULL)
-        I=new MediaInfoList;
+    if (I == NULL)
+    {
+        I = new MediaInfoList;
 
-    //Load GUI preferences
-    GUI_Configure();
+        //Load GUI preferences
+        GUI_Configure();
 
-    //File(s) in command line
-    #ifdef UNICODE
-        if (IsWin9X())
-        {
-            for (int I1=1; I1<=ParamCount(); I1++)
-                I->Open(ParamStr(I1).c_str());
-        }
-        else
-        {
-            int argc;
-            MediaInfoNameSpace::Char** argv=CommandLineToArgvW(GetCommandLineW(), &argc);
-            for (int I1=1; I1<argc; I1++)
-                I->Open(argv[I1]);
-        }
-    #else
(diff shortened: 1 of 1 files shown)


JeromeMartinez:
> PR https://github.com/MediaArea/MediaInfo/pull/851 and https://github.com/MediaArea/MediaInfo/pull/852 no longer needed.

They are already merged, please rebase and remove them in your PR.

cjee21:
Okay, dark mode switching should work well now and without resetting to basic view.

cjee21:
Removed commented-out testing code that I accidentally left there.

cjee21:
Oh no, HTML view is a little buggy now. Maybe need to put a refresh somewhere.

cjee21:
I can't figure out what is going on with HTML view now. It's a minor issue where HTML has to be clicked 2 times after switching dark mode else it is blank.

JeromeMartinez:
> It's a minor issue where HTML has to be clicked 2 times after switching dark mode else it is blank.

I can reproduce this issue but I'll survive with it for the moment :).

JeromeMartinez (approved):
Still not perfect but a lot better than before so I merge that.

cjee21:
An unrelated issue I found while testing is that if debug mode is toggled, the text view will be blank no matter what is toggled until application restart. I am not sure how debug mode is supposed to be used though.

JeromeMartinez:
General information: the code is very old (really... ~20 years) and messy, lot of unnecessary UI updates, we don't have much time on our side for it but if you are motivated we would be happy that you make it simpler.

Thank you again for your work.

JeromeMartinez:
>  if debug mode is toggled, the text view will be blank no matter what is toggled until application restart. I am not sure how debug mode is supposed to be used though.

Drag and drop again a file :).
It should analyse automatically the file with the new option but it was not implemented due to lack of time, not a big deal but it would be great it you update that.

cjee21:
> Thank you again for your work.
Glad to help.

> Drag and drop again a file :). It should analyse automatically the file with the new option but it was not implemented due to lack of time, not a big deal but it would be great it you update that.
Don't think I will do anything about that since I don't need that and no idea how to do. 😄

Noticed the contents of View -> Your system doesn't seem to correlate with the 3 radio button selections too but I don't use this page also.

PR #857, opened 2024-05-31
Windows GUI: Dark mode for HTML view

cjee21:
Inject CSS style into the HTML to give it a dark mode.

Screenshot 2024-05-31 171823

Let me know if you want to make changes to how it looks.

### DIFF
--- a/Source/GUI/VCL/GUI_Main.cpp
+++ b/Source/GUI/VCL/GUI_Main.cpp
@@ -110,6 +110,21 @@ bool __fastcall TMainF::WindowsDarkModeEnabled()
     return DarkModeEnabled;
 }
 
+//Function to inject a CSS style into html documents to make their look match the dark mode style
+std::wstring __fastcall TMainF::InjectDarkModeHTMLStyle(const wchar_t* HTMLDocument) {
+    const wchar_t* InsertionPoint = wcsstr(HTMLDocument, L"<head>");
+    const wchar_t* StyleContent = L"<style>body { background-color: #1F1F1F; color: #FFFFFF; }</style>";
+
+    if (InsertionPoint != nullptr) {
+        size_t InsertionPos = InsertionPoint - HTMLDocument + wcslen(L"<head>");
+        std::wstring ModifiedHTML(HTMLDocument);
+        ModifiedHTML.insert(InsertionPos, StyleContent);
+        return ModifiedHTML;
+    } else {
+        return HTMLDocument;
+    }
+}
+
 //---------------------------------------------------------------------------
 
 //***************************************************************************
@@ -460,7 +475,7 @@ void __fastcall TMainF::FormResize(TObject *Sender)
                 {
                     Page_Sheet_X_Web[KindOfStream]->Top =Page_Sheet_X[KindOfStream]->Top+1;
                     Page_Sheet_X_Web[KindOfStream]->Left=Page_Sheet_X[KindOfStream]->Width;
-				}
+                }
                 if (!Page_Sheet_X[KindOfStream+1]) //reached the bottom
                 {
                     //Bottom
@@ -904,6 +919,9 @@ void __fastcall TMainF::Refresh(TTabSheet *Page)
             //Creating file
             Ztring S1=I->Inform().c_str();
             File F;
+            if (M_Options_Darkmode->Checked) {
+                S1=InjectDarkModeHTMLStyle(I->Inform().c_str());
+            }
             if (FileName_Temp==__T(""))
             {
                 FileName_Temp=FileName::TempFileName_Create(__T("MI_"));
@@ -920,9 +938,12 @@ void __fastcall TMainF::Refresh(TTabSheet *Page)
         {
             Ztring TempA; TempA=Prefs->Translate(__T("At least one file"));
             Ztring Temp;
-            Temp+=L"about:<html><body>";
+            Temp+=L"about:<html><head></head><body>";
             Temp+=TempA.To_Unicode();
             Temp+=L"</body></html>";
+            if (M_Options_Darkmode->Checked) {
+                Temp=InjectDarkModeHTMLStyle(Temp.c_str());
+            }
             Page_HTML_HTML->Navigate((MediaInfoNameSpace::Char*)Temp.c_str());
         }
     }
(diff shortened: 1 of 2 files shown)


JeromeMartinez:
Great!
The scroll bar is still not in dark mode, any hope that it can be in dark mode too?

cjee21:
> Great! The scroll bar is still not in dark mode, any hope that it can be in dark mode too?

No idea for that one. This application is still using the old IE engine for HTML which isn't aware of dark mode as far as I know. There is a newer TEdgeBrowser for displaying HTML using the Edge engine but probably would not work on older Windows.

JeromeMartinez:
> There is a newer TEdgeBrowser for displaying HTML using the Edge engine but probably would not work on older Windows.

We dropped WinXP and our minimal requirement is Win7 now so I think it should be fine, if you are motivated go ahead on the switch.
If you don't want to risk to spend too much time on that, you may create a branch with a proof of concept only and we'll test it on a Win7 VM.

cjee21:
> > There is a newer TEdgeBrowser for displaying HTML using the Edge engine but probably would not work on older Windows.
> 
> We dropped WinXP and our minimal requirement is Win7 now so I think it should be fine, if you are motivated go ahead on the switch. If you don't want to risk to spend too much time on that, you may create a branch with a proof of concept only and we'll test it on a Win7 VM.

Windows 8.1 and below is dropped for this according to https://learn.microsoft.com/en-us/microsoft-edge/webview2/.
Also will require users to have Microsoft Edge WebView2 installed on their system if it is not already auto-installed.

JeromeMartinez:
> Also will require users to have Microsoft Edge WebView2 installed on their system if it is not already auto-installed.

[Win7 is still used by few % of people](https://gs.statcounter.com/windows-version-market-share/desktop/worldwide/#monthly-202405-202405-bar) despite it is EOL for a long time and it would not worth it to drop this support yet or bother people with an additional install, but if you don't mind we would like to use your skills when you are here, it would be great to have a patch ready and when we consider that it is time to switch we just merge it. We just don't promise about the time line.
And maybe we could update your PR by dynamically use one of the 2 engines depending on the OS, detection at runtime.

JeromeMartinez (approved):

JeromeMartinez:
(Note that this PR won't be in the next release as we already launched the build process, it will be for the one after the next one)

cjee21:
> [Win7 is still used by few % of people](https://gs.statcounter.com/windows-version-market-share/desktop/worldwide/#monthly-202405-202405-bar) despite it is EOL for a long time and it would not worth it to drop this support yet or bother people with an additional install, but if you don't mind we would like to use your skills when you are here, it would be great to have a patch ready and when we consider that it is time to switch we just merge it. We just don't promise about the time line. And maybe we could update your PR by dynamically use one of the 2 engines depending on the OS, detection at runtime.

Are there plans to use the Qt version for Windows? If so maybe better to spend effort on improving the Qt one instead. The latest Qt has native dark mode built-in for Windows without needing to do anything (looks and works better than this VCL implementation). By the way, the Windows 11 context menu if implemented can be used for any MediaInfo version on Windows. It is not dependent on this VCL version so it is fine to spend effort on that even if there are plans to retire the VCL version in the future.

JeromeMartinez:
> Are there plans to use the Qt version for Windows?

True

> It is not dependent on this VCL version so it is fine to spend effort on that even if there are plans to retire the VCL version in the future.

True too.

We are torn between polishing the VCL and switching to Qt at some point, but the gap between VCL and what is currently in Qt version is huge and we always postpone the full development.

cjee21:
I noticed MediaInfo left the temp HTML file on exit so I added code to delete that file on exit.

Note: We probably can use data URL and not need the temp file in the future when no longer need to support old browser engines.
https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/Data_URLs#browser_compatibility

cjee21:
TODO: Set WebView2 UDF to somewhere else it will probably crash when MediaInfo is installed in the read-only Program Files.
https://learn.microsoft.com/en-us/microsoft-edge/webview2/concepts/user-data-folder?tabs=win32
https://docwiki.embarcadero.com/Libraries/Sydney/en/Vcl.Edge.TCustomEdgeBrowser.UserDataFolder

cjee21:
Other changes required for implementing WebView2:

For installer:
- copy `C:\Program Files (x86)\Embarcadero\Studio\22.0\Redist\win32\WebView2Loader.dll` to `INSTDIR` on install
- delete `WebView2Loader.dll` from `INSTDIR` on uninstall
- `RMDir /r` `WebView2` from `INSTDIR` on uninstall
- `RMDir /r` `WebView2` from `APPDATA\MediaInfo` on uninstall
- might also want to remove the `MediaInfo\Plugin` and `MediaInfo` folders from `APPDATA` on uninstall. It is currently left there on uninstall.

For without installer:
- include `C:\Program Files (x86)\Embarcadero\Studio\22.0\Redist\win32\WebView2Loader.dll` in `.7z` archives

cjee21:
> We are torn between polishing the VCL and switching to Qt at some point, but the gap between VCL and what is currently in Qt version is huge and we always postpone the full development.

@JeromeMartinez Tried building Qt version using latest Qt (6.7.1) and latest MSVC (2022) just to see how it looks.

Here's how it looks on Windows 11:
Screenshot 2024-06-06 215704
Screenshot 2024-06-06 215732

JeromeMartinez:
> If it works on Windows 7 and 10 as well and you are satisfied with it, then I can make the pull request.

Please do open a PR, easier to track.

cjee21:
> We are torn between polishing the VCL and switching to Qt at some point, but the gap between VCL and what is currently in Qt version is huge and we always postpone the full development.

@JeromeMartinez I made some changes to the Qt version while testing. The changes are available here: https://github.com/cjee21/MediaInfo/tree/Qt. If you would like me to make a PR for this as well, just let me know.

This is how my branch currently looks on Windows 11 and Ubuntu:
Screenshot 2024-06-19 012642
Screenshot 2024-06-19 012718
Screenshot from 2024-06-19 01-35-38
Screenshot from 2024-06-19 01-35-56

Issue #858, opened 2024-05-31
24.05 Windows GUI broken on < Windows 10 Version 1803


nomakewan:
As of the latest update, MediaInfo no longer opens successfully on any version of Windows prior to Windows 10 Version 1803. This is because the API call to get the DPI info (added in #835 ) was only introduced into the Windows kernel as of that version (see [here](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-getsystemdpiforprocess)).

On versions of Windows prior to this, the installer will work just fine, but attempting to run MediaInfo itself will result in the following error:

![mediainfoerror](https://github.com/MediaArea/MediaInfo/assets/11485516/e1a5f8f7-a4f7-4814-9148-2f3eb98ce9c3)

The release page still says that MediaInfo supports Vista, 7, and 8 but as of 24.05 that is not the case. Of course the easiest solution would be to simply change the documentation to clearly state that versions of Windows prior to Windows 10 Version 1803 are no longer supported, but I thought it prudent to bring this to your attention.

Thank you!

JeromeMartinez:
Argh, we checked on Win7 that it was fine with Dark Mode, but we forgot to do same with High DPI support,
@cjee21, do you have an idea about how to have HigH DPI support while keeping support of older Windows?

cjee21:
> Argh, we checked on Win7 that it was fine with Dark Mode, but we forgot to do same with High DPI support, @cjee21, do you have an idea about how to have HigH DPI support while keeping support of older Windows?

Let me check. The high DPI support is mostly built into VCL framework, I did not implement API calls manually.

JeromeMartinez:
> I did not implement API calls manually.

In this case, it is.
`float DPIScale=static_cast<float>(GetSystemDpiForProcess(GetCurrentProcess()))/96;`

So there is a need to check the platform version before calling it.

cjee21:
> > I did not implement API calls manually.
> 
> In this case, it is. `float DPIScale=static_cast<float>(GetSystemDpiForProcess(GetCurrentProcess()))/96;`
> 
> So there is a need to check the platform version before calling it.

Oh this part. The window size part.

JeromeMartinez:
@cjee21 it is fine, I see that `GetVersionEx()` provides the Windows 10 version, I am adding a version check.

cjee21:
> @cjee21 it is fine, I see that `GetVersionEx()` provides the Windows 10 version, I am adding a version check.

[...]

PR #859, opened 2024-05-31
Windows GUI: Use older API for getting system DPI

cjee21:
Use older API to get system DPI for taking DPI scaling into account when setting window size so that it works on older Windows versions.

### DIFF
--- a/Source/GUI/VCL/GUI_Main.cpp
+++ b/Source/GUI/VCL/GUI_Main.cpp
@@ -243,7 +243,16 @@ __fastcall TMainF::TMainF(TComponent* Owner)
 void __fastcall TMainF::GUI_Configure()
 {
     //Hard coded
-    float DPIScale=static_cast<float>(GetSystemDpiForProcess(GetCurrentProcess()))/96;
+    OSVERSIONINFO osvi;
+    ZeroMemory(&osvi, sizeof(OSVERSIONINFO));
+    osvi.dwOSVersionInfoSize = sizeof(OSVERSIONINFO);
+    GetVersionEx(&osvi);
+    int DPI;
+    if (osvi.dwMajorVersion >= 10 && (osvi.dwMajorVersion > 10 || osvi.dwMinorVersion > 0 || osvi.dwBuildNumber >= 17134))
+        DPI=GetSystemDpiForProcess(GetCurrentProcess());
+    else
+        DPI=GetDeviceCaps(GetDC(NULL), LOGPIXELSX);
+    float DPIScale=static_cast<float>(DPI)/96;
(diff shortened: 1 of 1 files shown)


JeromeMartinez:
In theory `LOGPIXELSY` should be used for the height but I guess this is a corner case, but not blocking.

cjee21:
> In theory `LOGPIXELSY` should be used for the height but I guess this is a corner case, but not blocking.

Isn't X or Y having same DPI? There is only one DPI setting.

JeromeMartinez:
> Isn't X or Y having same DPI? There is only one DPI setting.

I bet that in practice they are same for 99.999% of devices but my understanding of the theory is that you can get 2 different values between width and height with this API.

JeromeMartinez:
> There is only one DPI setting.

With this API there are 2, `LOGPIXELSX` ("Number of pixels per logical inch along the screen width.") and `LOGPIXELSY` ("Number of pixels per logical inch along the screen height.")

cjee21:
> With this API there are 2, `LOGPIXELSX` ("Number of pixels per logical inch along the screen width.") and `LOGPIXELSY` ("Number of pixels per logical inch along the screen height.")

As far as I know, there is no way in Windows to set different DPI for X and Y. So I don't know if it is possible to have two different values for that API. Wonder why it has two though.

JeromeMartinez:
> Wonder why it has two though.

I bet that they expected to manage non square pixels (it is currently part of the past for TV only but it could return one day in computer screens... Or not) at some point and that it was abandoned in the meantime.

cjee21:
> I bet that they expected to manage non square pixels (it is currently part of the past for TV only but it could return one day in computer screens... Or not) at some point and that it was abandoned in the meantime.

Probably because it seems this API can be used for printers too. Maybe printers can have different X and Y.

cjee21:
Tested on a single monitor at 100 - 200% DPI scale and it works the same as when using the other API.

JeromeMartinez (changes requested):
Please update with https://github.com/MediaArea/MediaInfo/issues/858#issuecomment-2142177472 ASAP, then I merge and do a new public version.

cjee21:
> Please update with [#858 (comment)](https://github.com/MediaArea/MediaInfo/issues/858#issuecomment-2142177472) ASAP, then I merge and do a new public version.

Alright.

cjee21:
Argh the space and tab issue again. 😓

JeromeMartinez (approved):

JeromeMartinez:
@cjee21 FYI https://github.com/MediaArea/MediaInfo/pull/860#issuecomment-2142503257.

PR #860, opened 2024-05-31
Windows GUI: Use older API for getting system DPI, update

JeromeMartinez:


JeromeMartinez:
For an unknown reason the binary is trying to load the function even if the function is not called, this was not expected, usually the crash happens only when the function is called, so we remove it completely as an hotfix for the release.

cjee21:
> For an unknown reason the binary is trying to load the function even if the function is not called, this was not expected

Is the version check working properly? If it is then I wonder if this has anything to do with speculative execution.

cjee21:
> For an unknown reason the binary is trying to load the function even if the function is not called, this was not expected

Looks like it is looking for `GetSystemDpiForProcess` immediately on start before application code is run. Seems need to use `LoadLibrary` and `GetProcAddress` if we still want to use newer API on newer Windows.

cjee21:
@JeromeMartinez please take a look at #861

PR #861, opened 2024-06-01
Windows GUI: Use older API for getting system DPI, update2

cjee21:
Implement delayed loading of `user32.dll` and use `GetDpiForWindow`. Should work on Windows 7 to 11 now.

Tested on Windows 7 VM at 100% and 125%
Tested on Windows 11 PC at 100% and 150%

Nothing unexpected noticed, all seems fine.

### DIFF
--- a/Project/BCB/GUI/MediaInfo_GUI.cbproj
+++ b/Project/BCB/GUI/MediaInfo_GUI.cbproj
@@ -111,6 +111,7 @@
 			<Manifest_File>$(BDS)\bin\default_app.manifest</Manifest_File>
 			<AppDPIAwarenessMode>PerMonitorV2</AppDPIAwarenessMode>
 			<Icon_MainIcon>..\..\..\Source\Resource\Image\MediaInfo.ico</Icon_MainIcon>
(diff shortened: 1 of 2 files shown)


cjee21:
🤔 maybe should use `GetDpiForWindow` instead

cjee21:
Changed to `GetDpiForWindow`. Still works fine. Now new API works on Windows 10 Anniversary Update onwards.

JeromeMartinez:
I was not aware of `ILINK_DelayLoadDll`, I was implementing the dynamic load myself but this method is better.
`GetDpiForWindow` how many API is there for getting DPI??? :-p.

JeromeMartinez (approved):

cjee21:
> how many API is there for getting DPI??? :-p.

Many... https://learn.microsoft.com/en-us/windows/win32/hidpi/high-dpi-reference

JeromeMartinez:
> Many...

😱

JeromeMartinez:
Side note: I was "flooded" (ok, not so much, but still a couple within a couple of hours) by emails / reports about Win7 despite its EOL for several years, so Win7 support is still important.
(Me remembering some complains when we stopped WinXP support due to new Visual Studio not supporting it anymore few years ago so several years after its end of life... We'll have to support Win7 for several years)

PR #871, opened 2024-06-18
Windows GUI: Implement theme preferences

cjee21:
Enable user to save their preferences for application theme.

This is an implementation of the suggestion in https://github.com/MediaArea/MediaInfo/pull/836#issuecomment-2124305168 which I am only able to implement now after studying and figuring out how MediaInfo handle preferences.

Screenshot 2024-06-18 170430

Fixed in latest force push:
>Still needs more testing and to fix this bug that occurs when MediaInfo tries to start in dark mode:
Screenshot 2024-06-18 170453
This is something we ran into in the initial dark mode implementation. It is supposedly caused by changing the theme during form show. Before this, we worked around by setting the theme in the form creation part. However, we need to find another place for it now because we need to read the preferences to set the theme and can only do so after the preferences are initialized and loaded.

### DIFF
--- a/Source/Common/Preferences.cpp
+++ b/Source/Common/Preferences.cpp
@@ -258,6 +258,7 @@ int Preferences::Config_Save()
     if (Config(__T("Donated")).empty()) Config(__T("Donated"))=__T("0");
     if (Config(__T("Donate_Display")).empty()) Config(__T("Donate_Display"))=__T("1");
     if (Config(__T("Sponsored")).empty()) Config(__T("Sponsored"))=__T("0");
+    if (Config(__T("Theme")).empty()) Config(__T("Theme"))=__T("0");
 
     HANDLE Temp=CreateFile((BaseFolder+__T("MediaInfo.cfg")).c_str(), GENERIC_WRITE, 0, NULL, CREATE_ALWAYS, 0, NULL);
     if (Temp==INVALID_HANDLE_VALUE)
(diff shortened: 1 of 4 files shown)


JeromeMartinez:
@cjee21 thank you for your help there, very appreciated.

JeromeMartinez:
(and I know that this 20 year old code is super messy, it needs a bug clean up :( )

cjee21:
While looking around I came across this `TODO`:

```C++
	if (Page->ActivePage==Page_Sheet) //TODO : find WHY there is a crash without this...
	   ChangePage(Page_Easy);

	delete Prefs; Prefs=NULL;
	delete I; I=NULL;
```

Found that the crash is due to this line:
```C++
delete Prefs; Prefs=NULL;
```

cjee21:
While looking around I came across this `TODO`:

```C++
	if (Page->ActivePage==Page_Sheet) //TODO : find WHY there is a crash without this...
	   ChangePage(Page_Easy);

	delete Prefs; Prefs=NULL;
	delete I; I=NULL;
```

Found that the crash is due to this line:
```C++
delete Prefs; Prefs=NULL;
```

After debugging, found that after `delete Prefs; Prefs=NULL;`, it seems to still run `FormResize` then at `Total+=Prefs->Details[Prefs_Sheet](Z1, 4).To_int32s();`, the crash occur because Prefs is already deleted. Putting `delete Prefs; Prefs=NULL;` in `FormDestroy` does not work too as this also seems to run before the form is actually destroyed.

cjee21:
After moving stuff around, it appears to be working as expected now but probably need more testing to make sure nothing got accidentally broken.

JeromeMartinez:
@cjee21 please rebase then I do so tests on the GUI.

JeromeMartinez (approved):
It could also be in the main page of the preference window (below "output format") in another PR, if you are motivated.

Thank you again, your help is greatly appreciated!

PR #876, opened 2024-06-19
Qt GUI Improvements

cjee21:
A collection of improvements made while I was exploring the Qt version of MediaInfo.

- Use MSVC2022 for building the x86_64 Windows build
  - Changes to project file to successfully build using MSVC2022 including fixes for linker errors due to path issues
- Enable Control Flow Guard (CFG) for x86_64 Windows build
  - Enables CFG as [strongly encouraged by Microsoft](https://learn.microsoft.com/en-us/windows/win32/secbp/control-flow-guard).
  - For full protection, MediaInfoLib, ZenLib and zlib should be built with `/guard:cf` enabled in `Project | Properties | Configuration Properties | C/C++ | Code Generation` of their respective Visual Studio Projects.
- Reverts commit https://github.com/MediaArea/MediaInfo/commit/18b6e3ca7d52940c6b0a6a89aebd51672ee0db72
  - This commit it is not cross-platform and causes missing icons in both Windows and Ubuntu.
- Use newer 'Close' icon when available
  - Uses newer `QIcon::fromTheme(QIcon::ThemeIcon::WindowClose)` when built with Qt > 6.7
  - This newer icon is visible in dark mode on Windows unlike previous one.
- Prefer Cascadia Mono font when available
  - For systems with Cascadia Mono font (the default font for Windows Terminal) installed, it will be used else old font selection behaviour is maintained.
  - Cascadia Mono is more readable than the previously selected font on Windows and is also consistent with current VCL version of MediaInfo.
- Prevent toolbar from being moved
  - Previously, it was possible to detach the toolbar.
- Prevent clipped text in easy view
  - Make it resize properly to prevent clipped text in easy view when viewing certain files with multi-line metadata due to text wrap.
- Other improvements
  - Remove unused includes in `mainwindow.cpp`.
  - Fix inconsistent indents in `mainwindow.cpp`.

Tested with Qt 6.7.2 and MSVC2022.

**Windows 11 screenshots:**
Screenshot 2024-06-19 192245
_Process Explorer showing 64-bit Qt version with DEP, ASLR, CFG and Per-monitor DPI Awareness enabled_
Screenshot 2024-06-19 190541
Screenshot 2024-06-19 190557
Screenshot 2024-06-19 190618

**Ubuntu screenshots:**
Screenshot from 2024-06-19 19-10-41
Screenshot from 2024-06-19 19-10-58
Screenshot from 2024-06-19 19-11-10

**Previous screenshots for comparison:**
https://github.com/MediaArea/MediaInfo/pull/857#issuecomment-2152643509
https://github.com/MediaArea/MediaInfo/pull/857#issuecomment-2176644649

### DIFF
--- a/Project/QMake/GUI/MediaInfoQt.pro
+++ b/Project/QMake/GUI/MediaInfoQt.pro
@@ -188,7 +188,7 @@ win32 {
         }
     }
     QMAKE_CXXFLAGS += /guard:cf
-    QMAKE_LFLAGS += /guard:cf
+    QMAKE_LFLAGS += /guard:cf /CETCOMPAT
 }
 
 INCLUDEPATH += _Automated \
--- a/Project/QMake/GUI/MediaInfoQt.pro
+++ b/Project/QMake/GUI/MediaInfoQt.pro
@@ -185,6 +185,8 @@ win32 {
             } else {
                 error("zlib not found on system")
             }
+            QMAKE_CXXFLAGS += /guard:cf
+            QMAKE_LFLAGS += /guard:cf
         }
     }
 }
--- a/Project/QMake/GUI/MediaInfoQt.pro
+++ b/Project/QMake/GUI/MediaInfoQt.pro
@@ -73,21 +73,21 @@ win32 {
         INCLUDEPATH+=$$[QT_INSTALL_PREFIX]/include/QtCore/$$[QT_VERSION]
 
         contains(QT_ARCH, ARM) {
-            exists(../../../../MediaInfoLib/Project/MSVC2017/Arm/Release/MediaInfo-Static_UWP.lib) {
+            exists(../../../../MediaInfoLib/Project/MSVC2022/Arm/Release/MediaInfo-Static_UWP.lib) {
                 INCLUDEPATH += ../../../../MediaInfoLib/Source
-                LIBS += ../../../../MediaInfoLib/Project/MSVC2017/Arm/Release/MediaInfo-Static_UWP.lib
+                LIBS += ../../../../MediaInfoLib/Project/MSVC2022/Arm/Release/MediaInfo-Static_UWP.lib
             } else {
                 error("libmediainfo not found on system")
             }
 
-            exists(../../../../ZenLib/Project/MSVC2017/Arm/Release/ZenLib_UWP.lib) {
+            exists(../../../../ZenLib/Project/MSVC2022/Arm/Release/ZenLib_UWP.lib) {
                 INCLUDEPATH += ../../../../ZenLib/Source
-                LIBS += ../../../../ZenLib/Project/MSVC2017/Arm/Release/ZenLib_UWP.lib
+                LIBS += ../../../../ZenLib/Project/MSVC2022/Arm/Release/ZenLib_UWP.lib
             } else {
                 error("libzen not found on system")
             }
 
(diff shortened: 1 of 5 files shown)


JeromeMartinez (changes requested):
> A collection of improvements made while I was exploring the Qt version of MediaInfo.

Thank you so much, we didn't yet have the time to focus on such improvement and it is great that a third party uses the possibilities of open source for improving MediaInfo, and that you take the time to deeply test the results, we know that it is a lot of time.

> Use MSVC2022 for building the x86_64 Windows build 

I prefer to keep a coherence so to change that everywhere.
About `$$PWD`, what is the issue without it? It seems to work well without it.

> Enable Control Flow Guard (CFG) for x86_64 Windows build 

Beside Qt project, if you don't mind, please check, enable if needed, test and send another PR for VS2022 as well as VS2019 (we didn't yet upgraded VS on our build farm), for all repositories. Nothing mandatory if you don't want to do that, just a wish.

As we are there, you may be curious about [CET](https://techcommunity.microsoft.com/t5/windows-os-platform-blog/understanding-hardware-enforced-stack-protection/ba-p/1247815), and try [CET option](https://learn.microsoft.com/en-us/cpp/build/reference/cetcompat). As we don't have [a CPU supporting CET](https://www.epey.co.uk/cpu/e/YToxOntpOjUwOTc7YToxOntpOjA7czo3OiIyMDUyNDEzIjt9fV9iOjA7=/), we didn't try but it would be great if it is doable.

JeromeMartinez:
> I can change for others as well but it will be untested.

Similar update so I take the risk :).

> Without this I get many linker errors of things not found during the linking stage.

Weird, Fedora uses it and seems fine, @g-maxime can you check that `mediainfo-qt` package still compiles on Fedora?

> There is also [Spectre mitigations](https://learn.microsoft.com/en-us/cpp/build/reference/qspectre-load?view=msvc-170) but I'm not sure if relevant for MediaInfo. This one is known to affect performance.

IMO the "cost" of this option is higher than the benefit we may have with it, so I wouldn't add it until they find a less impacting mitigation.

cjee21:
> Weird, Fedora uses it and seems fine, @g-maxime can you check that `mediainfo-qt` package still compiles on Fedora?

Only affects Windows MSVC as far as I know. It built fine on Ubuntu without any changes.

cjee21:
> Beside Qt project, if you don't mind, please check, enable if needed, test and send another PR for VS2022 as well as VS2019 (we didn't yet upgraded VS on our build farm), for all repositories. Nothing mandatory if you don't want to do that, just a wish.

PRs created for MediaInfoLib, ZenLib and zlib.

JeromeMartinez (approved):

cjee21:
FYI The HTML view in Qt version already has proper dark mode support on Windows.

Screenshot 2024-06-20 124639

cjee21:
> PRs created for MediaInfoLib, ZenLib and zlib.
> 
> Also for MediaInfo CLI. There are probably other MSVC Projects which I did not touch because I do not know their purpose.

@JeromeMartinez 
Confirmed that most recent dev build has CFG and CET enabled for MediaInfo DLLs and CLI
Screenshot 2024-06-20 230358
Screenshot 2024-06-20 230420
Screenshot 2024-06-20 230500
Screenshot 2024-06-20 230620

Issue #868, opened 2024-06-17
open folder / limitation?


schimschewski:
i'm using very often the "open folder"-feature to export the entire movies in subdirectories e.g.
this was working like a charm in <= version 23.x

since 24.0.1 there seems to be a limitation or some cache problems, it does not show up all results. reopening the folder results in getting some missing files (sometimes only 5 and 10) from the first attempt.
since 24.0.5 it does not show up anything, seems like no recursive search in subdirectories is done.

it this a bug or an expected behaviour? see screenshots attached.

![mediainfo01](https://github.com/MediaArea/MediaInfo/assets/37181405/9cb43b5d-6009-46d8-b1ff-b74edf41e5b9)
![mediainfo02](https://github.com/MediaArea/MediaInfo/assets/37181405/6de0c2a0-c4c6-4867-8c0a-673827e7c864)
![mediainfo03](https://github.com/MediaArea/MediaInfo/assets/37181405/cede2f16-0d38-4909-8d97-9524623829f2)
![mediainfo04](https://github.com/MediaArea/MediaInfo/assets/37181405/5d5156f1-7aab-494e-87bf-51d98add9c49)

JeromeMartinez:
> it this a bug or an expected behaviour?

Bug, but we can not reproduce it, subdirs are still well scanned on our side
What is the memory usage (through the task manager) during the scan?

By the way, using the command line version may be better for such slow task ;-).

schimschewski:
> > it this a bug or an expected behaviour?
> 
> Bug, but we can not reproduce it, subdirs are still well scanned on our side What is the memory usage (through the task manager) during the scan?
> 
> By the way, using the command line version may be better for such slow task ;-).

You mean the memory usage for the latest version?

![image](https://github.com/MediaArea/MediaInfo/assets/37181405/19b6e3c9-b301-4d57-907d-4959a8e3b133)

Talking for version 24.05.1:
The cpu takes some usage for only 3-4 seconds, and then it goes down. 
the memory usage is not affected. even after minutes -> no results shown
Its working for other directories with lower subdirs count very well, but stopped working for higher amount.
the same directory is still working with v23.

thanks for the hint with the cli. will give it a try

JeromeMartinez:
> Its working for other directories with lower subdirs count very well, but stopped working for higher amount.

Could you spot a minimum directory structure that demonstrate such issue?
We would like to be able to reproduce it on our side for easier debugging.

> the same directory is still working with v23.

[...]

Issue #788, opened 2024-01-11
How to properly build GUI on Windows?


DanH-ER:
When I build the Visual Studio project on Windows the resulting GUI app has the sheets and tree views disabled, is missing the preferences window, and just generally looks different that the official release from the website. So my question is how do I build the same UI that's in the official release? Is there a config or build flag step I'm missing?

The instructions in the readme only describe how to build Mac and Linux, not Windows.

DanH-ER:
I downloaded the community edition of Embarcadero and I can build the UI in 32 bit mode but when I try to build the 64 bit version I get errors similar to this...

[ilink64 Error] Error: Unresolved external 'ZenLib::ZtringListListF::ZtringListListF()' referenced from C:\REPOS\MEDIAINFO\PROJECT\BCB\GUI\WIN64\RELEASE\PREFERENCES.O

I built the BCB ZenLib project first, in both 32 and 64 bit, so I'm not sure what is causing this. 

Also I can't seem to build MediaInfoLib at all. The 32 bit version throws errors complaining about int128u and the the 64 bit version complains about __uuifof. 

Since you guys use this for the official build I assume you don't have these errors. Any tips on what might be causing them?

DanH-ER:
Never mind I see that the release script uses the 32bit GUI and the DLLs from Visual Studio. I think I see how it works.

cjee21:
Documenting how I managed to build the Windows GUI here to help others and for my future reference. Thanks to https://github.com/MediaArea/MediaInfo/issues/788#issuecomment-1887569305 and https://github.com/MediaArea/MediaInfo/issues/788#issuecomment-1888203511 for giving me an idea.

### How to build Windows GUI
1. Install the following required software on a Windows PC:
   - C++Builder Community Edition (https://www.embarcadero.com/products/cbuilder/starter)
   - Git for Windows (https://gitforwindows.org/)
2. Create an empty folder. We'll call this `MediaInfoDevelopment` for this guide.
3. Clone the following MediaInfo dependencies to `MediaInfoDevelopment` using Git:
   - https://github.com/MediaArea/MediaInfoLib
   - https://github.com/MediaArea/ZenLib
   - https://github.com/MediaArea/zlib
4. Clone MediaInfo (this repository) to `MediaInfoDevelopment`.
   - If you intend to contribute to MediaInfo, fork the repository on GitHub first and clone your fork instead.
5. The `MediaInfoDevelopment` folder should now have the following folders:
[...]

Issue #575, opened 2021-11-04
Out of memory after processing about 60k files with MediaInfo GUI


dust0vich:
Hello,
I am trying to do a report of about 700,000 files with the MediaInfo GUI to export to a CSV file as a custom sheet. It gets stuck with a out of memory error at around 60,000 files. I have plenty of memory allocated as well as am running the 64 bit version so there shouldn't be any restrictions on memory usage.

I was trying with the CLI version, but it doesn't seem to accept my sheet template that works in the GUI version and just spits out the default file info every time, also whenever I use wildcards it gives me nothing. 

Windows Server 2012 R2 with 20 GB ram
MediaInfo CLI/GUI 21.09

JeromeMartinez:
> I have plenty of memory allocated as well as am running the 64 bit version so there shouldn't be any restrictions on memory usage.

Actually there is one: the 64-bit installer installs the 64-bit Win explorer plugin but the GUI itself is still 32-bit (we didn't yet move to 64-bit for the GUI, planned but lack of time for working on that)

> I was trying with the CLI version

The CLI is definitely the way to go.

> but it doesn't seem to accept my sheet template that works in the GUI version

They both use the same engine. Possible to share (public or private) the template?
"sheet template"? AFAIK the sheet part has not exportable template and in practice it would be a mess in the UI to have so many files, so the idea here is to get the "custom template" in order to have a text file after that, then you do what you want with this text file.

dust0vich:
Here is the template:
[Duration.csv](https://github.com/MediaArea/MediaInfo/files/7478248/Duration.csv)
Generated by the GUI: 
![image](https://user-images.githubusercontent.com/17212092/140414436-1dddcd4b-441b-449a-b9bd-dffa32db262a.png)

Using other command line:
`mediainfo --Output=file://F:\Duration.csv "f:\Recordings\100\2019-03-01\Call Recording\blahblah.mp3"`
`mediainfo --Inform=file://F:\Duration.csv "f:\Recordings\100\2019-03-01\Call Recording\blahblah.mp3"`
just gives me the general info as if it was not specified.

JeromeMartinez:
> Here is the template:

OK, sheet template is only for GUI as it is "just" a list of fields and a size for GUI only

Corresponding CLI template:
```
mediainfo ../Downloads/MDV00428_example.dv "--Output=General;%FolderName%,%F
ileName%,%FileExtension%,%Duration/String5%\r\n"
```
[...]

PR #882, opened 2024-06-22
Windows GUI: Enable 64-bit version

cjee21:
Screenshot 2024-06-22 180755
Screenshot 2024-06-22 180653

Requires building 64-bit versions of ZenLib and zlib as well. For these two, no changes required other than enabling 64-bit.

Advantages of 64-bit version:
- Better performance?
  - 05 seconds 988 milliseconds vs 09 seconds 258 milliseconds on a single folder test
- Large address aware
- High Entropy VA / High-entropy 64-bit address space layout randomization (ASLR)

Issues that may be related: #575 #788 #868

### DIFF
--- a/Project/BCB/GUI/MediaInfo_GUI.cbproj
+++ b/Project/BCB/GUI/MediaInfo_GUI.cbproj
@@ -139,7 +139,7 @@
         <Manifest_File>$(BDS)\bin\default_app.manifest</Manifest_File>
         <Icon_MainIcon>..\..\..\Source\Resource\Image\MediaInfo.ico</Icon_MainIcon>
         <BT_BuildType>Debug</BT_BuildType>
-        <Defines>MEDIAINFODLL_NAME=L&quot;MediaInfo_i386.dll&quot;;$(Defines)</Defines>
+        <Defines>MEDIAINFODLL_NAME=L&quot;MediaInfo.dll&quot;;$(Defines)</Defines>
         <ILINK_DelayLoadDll>user32.dll;$(ILINK_DelayLoadDll)</ILINK_DelayLoadDll>
     </PropertyGroup>
     <PropertyGroup Condition="'$(Cfg_2_Win64)'!=''">
--- a/Project/BCB/GUI/MediaInfo_GUI.cbproj
+++ b/Project/BCB/GUI/MediaInfo_GUI.cbproj
@@ -1,15 +1,14 @@
 ﻿<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
     <PropertyGroup>
         <ProjectGuid>{8AF87745-B671-4B4E-A428-B47B72F174BA}</ProjectGuid>
-        <ProjectVersion>19.5</ProjectVersion>
+        <ProjectVersion>19.7</ProjectVersion>
         <FrameworkType>VCL</FrameworkType>
         <MainSource>MediaInfo_GUI.cpp</MainSource>
         <Base>True</Base>
         <Config Condition="'$(Config)'==''">Release</Config>
-        <Platform Condition="'$(Platform)'==''">Win32</Platform>
-        <TargetedPlatforms>1</TargetedPlatforms>
+        <Platform Condition="'$(Platform)'==''">Win64</Platform>
+        <TargetedPlatforms>3</TargetedPlatforms>
         <AppType>Application</AppType>
-        <CC_Suffix Condition="'$(CC_Suffix)'==''">c</CC_Suffix>
     </PropertyGroup>
     <PropertyGroup Condition="'$(Config)'=='Base' or '$(Base)'!=''">
         <Base>true</Base>
@@ -59,8 +58,8 @@
         <Base>true</Base>
     </PropertyGroup>
     <PropertyGroup Condition="'$(Base)'!=''">
-        <Defines>MEDIAINFO_DLL_RUNTIME;MEDIAINFODLL_NAME=L&quot;MediaInfo_i386.dll&quot;;$(Defines)</Defines>
-        <VerInfo_Keys>CompanyName=;FileDescription=;FileVersion=24.06.0.0;InternalName=;LegalCopyright=;LegalTrademarks=;OriginalFilename=;ProductName=;ProductVersion=;Comments=</VerInfo_Keys>
+        <Defines>MEDIAINFO_DLL_RUNTIME;$(Defines)</Defines>
+        <VerInfo_Keys>CompanyName=MediaArea.net;FileDescription=MediaInfo;FileVersion=24.06.0.0;InternalName=;LegalCopyright=;LegalTrademarks=;OriginalFilename=;ProductName=MediaInfo;ProductVersion=24.06.0.0;Comments=</VerInfo_Keys>
         <Manifest_File>None</Manifest_File>
         <VerInfo_IncludeVerInfo>true</VerInfo_IncludeVerInfo>
         <VerInfo_MinorVer>05</VerInfo_MinorVer>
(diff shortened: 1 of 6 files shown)


cjee21:
Need to clean up the project file and change paths to relative...

cjee21:
Please see the updated first post.

I recommend having 64-bit installer version due to the advantages listed in first post, the fact that most PCs should be 64-bit already and the effect of smaller installed size on 64-bit PCs due to removal of `MediaInfo_i386.dll`.

JeromeMartinez:
Not tested but my understanding is that the 32-bit binaries are no more in the installer, size is smaller so it is good, but the 32-bit related code won't work.
I don't wish to have both architectures in the installer at long term but it could be a good first step for this PR, please update the PR for embedding both architectures, then we'll work on another PR for having a 32-bit only installer and update the main installer for downloading this 32-bit installer when 32-bit is detected.

We'll have to update build scripts in https://github.com/MediaArea/MediaArea-Utils/ , please some patience, it is on our side but we have other things to finish before managing that (including building the GUI twice). It won't be forever but not in June.

Thank you for pushing us in that direction, just that in the cases of EdgeView and 64-bit we are the blockers and we need some time. But don't give up, we are still very happy with you pushing us.

> most PCs should be 64-bit

You can not imagine the count of emails I received when we stop WinXP support despite the fact it was end of life for years, I bet the same for 32-bit, actually even worse (we have no stat, sadly).

cjee21:
> Not tested but my understanding is that the 32-bit binaries are no more in the installer, size is smaller so it is good, but the 32-bit related code won't work. I don't wish to have both architectures in the installer at long term but it could be a good first step for this PR, please update the PR for embedding both architectures, then we'll work on another PR for having a 32-bit only installer and update the main installer for downloading this 32-bit installer when 32-bit is detected.

This PR makes the installer generate a universal one. It will use only 64-bit binaries on 64-bit PCs. For 32-bit PCs should be still the same as now unless I made a mistake somewhere.
On your side you just need to make sure to build 64-bit versions of ZenLib, zlib and CURL. Then for MediaInfo, just select 64-bit target and build with the updated project file in this PR.

> we have no stat, sadly

Found outdated stats: https://blogs.windows.com/windowsexperience/2010/07/08/64-bit-momentum-surges-with-windows-7/

JeromeMartinez:
> most PCs should be 64-bit

Now I have doubts, looks like [that 32-bit Windows is less than 1%](https://www.pcbenchmarks.net/os-marketshare.html) so I would think to remove it...

cjee21:
> > most PCs should be 64-bit
> 
> Now I have doubts, looks like [that 32-bit Windows is less than 1%](https://www.pcbenchmarks.net/os-marketshare.html) so I would think to remove it...

Well the last time I remember seeing a 32-bit Windows PC was several years ago on a low-end small-sized Intel tablet.

JeromeMartinez:
>  unless I made a mistake somewhere.

I have read it again and it seems that my first read was wrong.

> On your side you just need to make sure to build 64-bit versions of ZenLib, zlib and CURL. Then for MediaInfo, just select 64-bit target and build with the updated project file in this PR.

We need to find where it is in our scripts and add the right line for building the 64-bit version. Theses scripts are very old so we need to read and understand them again :-D.

cjee21:
> We need to find where it is in our scripts and add the right line for building the 64-bit version. Theses scripts are very old so we need to read and understand them again :-D.

Where I think it is after a quick look: https://github.com/MediaArea/MediaArea-Utils/blob/master/build_release/BuildRelease.bat#L285

cjee21:
> > most PCs should be 64-bit
> 
> Now I have doubts, looks like [that 32-bit Windows is less than 1%](https://www.pcbenchmarks.net/os-marketshare.html) so I would think to remove it...

Less than 0.1% if we look at the 64 vs 32 chart. That link also shows Windows 7 is less than 0.5% in the versions chart.

Here is another source of stats: https://store.steampowered.com/hwsurvey

If we assume these stats accurately represent the actual user base and also assume that all unlisted versions are 32-bit, then the percentage of 32-bit Windows will be around 3%.

Side note: take a look at the Windows 11 share in both sources... it's time for Windows 11 explorer context menu 😉

cjee21:
Corrected path for x64 CURL since I just saw its location in the repos.

> we'll work on another PR for having a 32-bit only installer 

Added as an additional commit.

> update the main installer for downloading this 32-bit installer when 32-bit is detected.

This one leave it up to your team since need server URL and I'm not good enough with NSIS to do this.

From another page:
> Package size is a hot topic for us, most people don't care at all and download GB of content but some others complain loudly when there is 1 more MB in the main package...

Then you might want to update your 8-year old 7-Zip that I came across at https://github.com/MediaArea/MediaArea-Utils-Binaries/tree/master/Windows/7-Zip. I tried compressing the extracted 7z dev snapshot with my latest 7-Zip using the same command line as your script and it resulted in a slightly smaller file size.

cjee21:
Improved the installer.

The first change is adding back the option to run MediaInfo after installation has completed. I saw that it was removed "because it is run in admin privileges". It is now run at user privileges.

The second change is regarding this:

> @JeromeMartinez Might want to add check to see if MediaInfo is running during uninstall or add `/REBOOTOK` switch as currently it appears to complete uninstall but leave behind MediaInfo.exe and the DLL if MediaInfo is running during uninstall.

Now if MediaInfo is running or something is using `MediaInfo.dll` during uninstall, it will now move the in-use files to `%TEMP%` and schedule them for deletion on next reboot. It will then ask if the user want to reboot immediately or later on completion. The `Program Files\MediaInfo` folder will be removed cleanly unless there is still something there. If the user defers reboot and re-installs MediaInfo, the new installation will not be affected after reboot since the old files that are scheduled for deletion has been moved/renamed.

If MediaInfo is not running and nothing is using `MediaInfo.dll` during uninstall, there is no change in behaviour. The `Program Files\MediaInfo` folder will be removed cleanly unless there is still something there. There will be no reboot prompt.

This is implemented using macros from `Library.nsh`. The macros can possibly be used to handle installation of all exe and dlls and registration/unregistration/removal of `MediaInfo_InfoTip.dll` as well but I did not attempt this to keep changes to a minimal to avoid breaking something.

cjee21:
Add 64-bit only installer.

Size comparison:

```
MediaInfo\Release>dir | find "MediaInfo_GUI"
27/06/2024  11:20 AM         9,887,983 MediaInfo_GUI_24.05_Windows.exe
27/06/2024  11:20 AM         4,973,713 MediaInfo_GUI_24.05_Windows_i386.exe
27/06/2024  11:20 AM         5,521,995 MediaInfo_GUI_24.05_Windows_x64.exe
```

Maybe can do something like this if want to minimize download size:
https://firefox-source-docs.mozilla.org/browser/installer/windows/installer/StubInstaller.html

cjee21:
Re-based on 24.06 while ensuring version numbers in project file are updated. Also reduced number of commits and moved an unrelated commit to another branch where that file was changed over there.

JeromeMartinez:
> Maybe can do something like this if want to minimize download size:

The Stub installer is also something we are thinking to add while changing to a 64-bit offer so we show 1 small installer only to all download websites without issue with 32-bit machines (there may actually be more than estimated in the wild).

cjee21:
Rebased and changed 24.05 to 24.06 in installer scripts that I missed.

JeromeMartinez:
For reference https://github.com/MediaArea/MediaArea-Utils/pull/297 and https://github.com/MediaArea/MediaInfo/pull/883 were required for having everything ready for our workflow.

JeromeMartinez:
[dev snapshots](https://mediaarea.net/download/snapshots/binary/mediainfo-gui/20240702-4/)

cjee21:
> [dev snapshots](https://mediaarea.net/download/snapshots/binary/mediainfo-gui/20240702-4/)

This build is not from latest master and still has disappearing menu glitch but it's okay, I found a newer dev snapshot. I will use it until the next version to enjoy the new changes and also as additional testing.

Might want to update the project files for ZenLib and zlib to add 64-bit targets so that people who use IDE to build will not have trouble. They can easily manually add the target themselves though. I didn't expect building from CLI to work without 64-bit target in project file but it does.

cjee21:
First bug discovered. If open a file from explorer context menu then immediately switch from Basic to Graph view, this pops up:
Screenshot 2024-07-03 120730

If switch to HTML view then to Graph view, it works.

Issue #891, opened 2024-06-28
24.06 - Open Preferences and click on "Adavanced" -> GUI buttons appear


Klaus1189:
Open MediaInfo
Open "Preferences" and click on "Advanced" then despite search is disabled and version is up to date buttons appear:
https://drive.google.com/file/d/1PsJd_SAZsLTQygCpNOEmkTpY_UubWRFg/view?usp=sharing

And despite "toolbar" is disabled after closing the "Preferences" the toolbar is not hidden, only the buttons disappear but the place stays like it is still shown, but on only a cut off non-working orange button is shown is this area:
https://drive.google.com/file/d/1PsJd_SAZsLTQygCpNOEmkTpY_UubWRFg/view?usp=sharing

cjee21:
There's another minor 'Preferences' window bug I found. When clicking 'More...' next to 'Language' on 'Setup' page, it goes to 'Customize' -> 'Graph'. I'm guessing it should be going to 'Customize' -> 'Language'.

cjee21:
> And despite "toolbar" is disabled after closing the "Preferences" the toolbar is not hidden, only the buttons disappear but the place stays like it is still shown, but on only a cut off non-working orange button is shown is this area: https://drive.google.com/file/d/1AamHX_5Ql0HlsKfBwImyLM35kywfcHOE/view?usp=sharing

Steps to reproduce
1. Open 'Preferences' -> 'Advanced'
2. Disable 'Show toolbar' and click 'OK' 
3. Do step 1 again
4. Click 'OK' without touching anything else

Klaus1189:
If you need anything in this issue, please let me know. Testing or anything, I do it for you. 

And translation later, it will take some time. For RTSS and MSI Afterburner I also needed some time for full new translation:
https://forums.guru3d.com/threads/german-translation.447957/

cjee21:
I wonder why many things seem to happen on the main window upon clicking advanced in preferences window. I don't see anything that may cause that in this function: https://github.com/MediaArea/MediaInfo/blob/dda61d6a99aa58fbf69b488669c2ac26242222b2/Source/GUI/VCL/GUI_Preferences.cpp#L629

cjee21:
Debugging results:

[...]

PR #902, opened 2024-07-02
Windows GUI: Code clean-up, organization and fixes

cjee21:
- Remove unused donate buttons and related code
  - As mentioned in https://github.com/MediaArea/MediaInfo/issues/891#issuecomment-2200575320
- Organize `GUI_Main.cpp`
  - Move functions to the appropriate sections and improve some comments for easier finding and understanding in the future
- Upgrade open file dialog
  - Use the same dialog type for opening files as the one for opening folders
  - Should not have visible difference since old dialog is auto upgraded by VCL at runtime to new one when running on Windows Vista+
  - This change is for consistency with folder open dialog and maybe make it more efficient by dropping support of old dialog
  - The new dialogs work on Windows Vista+
- Partial fix for 2nd issue of #891 
  - The GUI is now properly displayed following user preferences when Preferences window is closed
  - The issues that happen when Preferences->Advanced is opened remains
- Fix the file selector button in Export window is too big at high-DPI
  - Similar to recent changes for buttons in Main and Preferences windows


JeromeMartinez:
Thank you for the clean up!

Issue #203, opened 2017-11-19
Ability to remember maximize/size of window and type of view


deniskx:
Please create an ability to remember maximize and size of window and type of view "View-Text/HTML".
I'm tired of changing every time.
Forgive my cry of the soul, but I think you are using your monopoly position (there is no alternative to MediaInfo) so as not to improve graphical usability.

JeromeMartinez:
> your monopoly position

I have difficulties to understand that, please argue about how permitting 7+ billions of persons [to use, copy, modify, distribute](https://mediaarea.net/en/MediaInfo/License), or [send patchs](https://github.com/MediaArea/MediaInfo/pulls)... has a link with using a monopoly position.

As it looks like that you think that working for free is the best thing on earth, I am interested in assigning you several tasks, for $0 obviously (you think that this rate is fair if I understand well your message).

> I'm tired of changing every time

I am tired to get aggressive complains like these ones from people who paid $0 for a tool and/or support contract, especially because I bet that I would get a "no way, I am not your slave" if I ask such person to work for me for free.

***

More seriously, you are 100% free to submit a patch (made by you or someone else, your task is just to find this person) so your sentence is false. If you think it is a feature you really need, we'll not "use our monopoly position" for blocking your patch, go ahead.

Issue here is that for the moment nobody was interested enough in this feature for developing it or sponsoring it alone; as this feature is requested by several people but not up to paying it alone (I hope that if they ask for that, they'd sponsor a bit it), we'll open the possibility to have mini-crowdfunding for such feature [soon](https://mediaarea.net/SupportUs/FAQ#votes) via [memberships](https://mediaarea.net/SupportUs/Individual), as it looks like it is a very important feature for you I have no doubt that you'll participate.

Eld0r:
I couldn't have said it better.
I miss it too, but I optimized my custom view. Of course it concerns all views.
So sb. need to pay for it. I would program it on my own, but I have no free time.
I must use my time not studying, for work to pay my Studium.
NeedSponsor is a understandable, suitable Label for feature requests 👍

regs01:
May I suggest a workaround. When in Tree or Text mode make default height as 850 * monitor scale (850 in 100%, 1275 in 150%, 1700 in 200% etc), but no more than 90% of monitor height, so would looking good on large screen and fit on small screens as well.

Wouldn't require to save anything, but still would be better.

garfnodie:
Bump.

I ask again, can I pay for a version with more customizability?

PR #903, opened 2024-07-02
Save window position and dimensions. Improve Preferences Form.

regs01:
Added optional ability to save window position and/or dimensions.

Also improved Preferences form a bit. Everything is aligned now. And it's resizeable now, which is more convenient.

### DIFF
--- a/Source/Common/Preferences.cpp
+++ b/Source/Common/Preferences.cpp
@@ -251,6 +251,8 @@ int Preferences::Config_Save()
     if (Config(__T("ShellExtension")).empty()) Config(__T("ShellExtension"))=__T("1");
     if (Config(__T("ShellExtension_Folder")).empty()) Config(__T("ShellExtension_Folder"))=__T("0");
     if (Config(__T("ShellInfoTip")).empty()) Config(__T("ShellInfoTip"))=__T("0");
+	if (Config(__T("RememberWindowPosition")).empty()) Config(__T("RememberWindowPosition"))=__T("0");
+	if (Config(__T("RememberWindowDimensions")).empty()) Config(__T("RememberWindowDimensions"))=__T("1");
     if (Config(__T("ShowToolBar")).empty()) Config(__T("ShowToolBar"))=__T("1");
     if (Config(__T("ShowMenu")).empty()) Config(__T("ShowMenu"))=__T("1");
     if (Config(__T("CloseAllAuto")).empty()) Config(__T("CloseAllAuto"))=__T("1");
--- a/Source/GUI/VCL/GUI_Main.cpp
+++ b/Source/GUI/VCL/GUI_Main.cpp
@@ -6,6 +6,7 @@
 
 //---------------------------------------------------------------------------
 #include <vcl.h>
+#include <Math.hpp>
 #pragma hdrstop
 #include "GUI/VCL/GUI_Main.h"
 #ifndef MEDIAINFOGUI_PREFS_NO
@@ -416,42 +417,88 @@ void __fastcall TMainF::GUI_Configure()
 
     //Configure theme
     ConfigTheme();
-}
 
-//---------------------------------------------------------------------------
-void __fastcall TMainF::FormShow(TObject *Sender)
-{
-    //Set window size and position
-    OSVERSIONINFO osvi;
-    ZeroMemory(&osvi, sizeof(OSVERSIONINFO));
-    osvi.dwOSVersionInfoSize = sizeof(OSVERSIONINFO);
-    GetVersionEx(&osvi);
-    int DPI;
-    if (osvi.dwMajorVersion >= 10 && (osvi.dwMajorVersion > 10 || osvi.dwMinorVersion > 0 || osvi.dwBuildNumber >= 14939))
-        DPI=GetDpiForWindow(WindowHandle);
-    else
-        DPI=GetDeviceCaps(GetDC(NULL), LOGPIXELSX);
-    float DPIScale=static_cast<float>(DPI)/96;
-    float ScaledScreenWidth=Screen->Width/DPIScale;
-    float ScaledScreenHeight=Screen->Height/DPIScale;
-    Width=500;
-    Height=400;
-    if (ScaledScreenWidth>=1024)
-        Width=700;
-    if (ScaledScreenWidth>=1280)
-        Width=830;
-    if (ScaledScreenHeight>=768)
-        Height=500;
-    if (ScaledScreenHeight>=1024)
-        Height=600;
-    Width*=DPIScale;
-    Height*=DPIScale;
-    Left=(Screen->Width-Width)/2;
-    Top=(Screen->Height-Height)/2;
+	//Set window size and position
+	// todo: move all property assignments into FormCreate
+	// note: VCL version? It's been having built-in DPI support since long ago. And it should be Monitor instead of Screen.
+	// OSVERSIONINFO osvi;
+	// ZeroMemory(&osvi, sizeof(OSVERSIONINFO));
+	// osvi.dwOSVersionInfoSize = sizeof(OSVERSIONINFO);
+	// GetVersionEx(&osvi);
+	// int DPI;
+	// if (osvi.dwMajorVersion >= 10 && (osvi.dwMajorVersion > 10 || osvi.dwMinorVersion > 0 || osvi.dwBuildNumber >= 14939))
+	//	   DPI=GetDpiForWindow(WindowHandle);
+	// else
+	// 	   DPI=GetDeviceCaps(GetDC(NULL), LOGPIXELSX);
+	// float DPIScale=static_cast<float>(DPI)/96;
+	//
+	// float fpUnscaledMonitorWidth=Monitor->Width/DPIScale;
+	// float fpUnscaledMonitorHeight=Monitor->Height/DPIScale;
 
-    //Refresh global
-    FormResize(NULL);
-    Refresh();
+	if (Prefs->Config(__T("RememberWindowDimensions")) == __T("1")) {
+
+		int nPrefWidth = Width;
+		int nPrefHeight = Height;
+
+		if (Prefs->Config(__T("FormWidth")).IsNumber())
+			nPrefWidth = Prefs->Config(__T("FormWidth")).To_int32s();
+		if (Prefs->Config(__T("FormHeight")).IsNumber())
+			nPrefHeight = Prefs->Config(__T("FormHeight")).To_int32s();
+
+		nPrefWidth = Min(Monitor->WorkareaRect.Width(), nPrefWidth);
+		nPrefHeight = Min(Monitor->WorkareaRect.Height(), nPrefHeight);
+
+		SetBounds(Left, Top, nPrefWidth, nPrefHeight);
+
+	}
+	else {
+
+		// using VCL
+		float fpUnscaledMonitorWidth = Monitor->Width / ScaleFactor;
+		float fpUnscaledMonitorHeight = Monitor->Height / ScaleFactor;
+
+		Width=500;
+		Height=400;
+		if (fpUnscaledMonitorWidth>=1024)
+			Width=700;
+		if (fpUnscaledMonitorWidth>=1280)
+			Width=830;
+		if (fpUnscaledMonitorHeight>=768)
+			Height=500;
+		if (fpUnscaledMonitorHeight>=1024)
+			Height=600;
+		if (fpUnscaledMonitorHeight>=1440)
+			Height=900;
+		Width *= ScaleFactor;
+		Height *= ScaleFactor;
+		// note: just use poScreenCenter
+		// Left=(Screen->Width-Width)/2;
+		// Top=(Screen->Height-Height)/2;
+
+	}
+	if (Prefs->Config(__T("RememberWindowPosition")) == __T("1")) {
+
+		int nPrefTop = 0;
+		int nPrefLeft = 0;
+
+		if (Prefs->Config(__T("FormTop")).IsNumber())
+			nPrefTop = Prefs->Config(__T("FormTop")).To_int32s();
+		if (Prefs->Config(__T("FormLeft")).IsNumber())
+			nPrefLeft = Prefs->Config(__T("FormLeft")).To_int32s();
+
+		nPrefTop = Max(nPrefTop, 0);
+		nPrefTop = Min(nPrefTop,  Monitor->WorkareaRect.Height() - Height);
+		nPrefLeft = Max(nPrefLeft, 0);
+		nPrefLeft = Min(nPrefLeft, Monitor->WorkareaRect.Width() - Width);
+		SetBounds(nPrefLeft, nPrefTop, Width, Height);
+
+		Position = poDesigned;
+
+	}
+	else {
+		Position = poScreenCenter;
+	}
+
 }
 
 //---------------------------------------------------------------------------
@@ -460,6 +507,21 @@ void __fastcall TMainF::FormClose(TObject *Sender, TCloseAction &Action)
     //Delete temp HTML file on close
     if (FileName_Temp!=__T(""))
         File::Delete(FileName_Temp);
+
+	bool bCallPrefsSave = false;
+	if (Prefs->Config(__T("RememberWindowPosition")) == __T("1")) {
+		Prefs->Config(__T("FormTop")).From_Number((int)this->Top);
+		Prefs->Config(__T("FormLeft")).From_Number((int)this->Left);
+		bCallPrefsSave = true;
+	}
+	if (Prefs->Config(__T("RememberWindowDimensions")) == __T("1")) {
+		Prefs->Config(__T("FormWidth")).From_Number((int)this->Width);
+		Prefs->Config(__T("FormHeight")).From_Number((int)this->Height);
+		bCallPrefsSave = true;
+	}
+
+	if (bCallPrefsSave)
+		Prefs->Config.Save();
 
     //The form is closed and all allocated memory for the form is freed.
     Action = caFree;
--- a/Source/GUI/VCL/GUI_Main.dfm
+++ b/Source/GUI/VCL/GUI_Main.dfm
@@ -14,7 +14,6 @@ object MainF: TMainF
   OnClose = FormClose
   OnDestroy = FormDestroy
   OnResize = FormResize
-  OnShow = FormShow
   TextHeight = 13
   object SpeedButton1: TSpeedButton
     Left = 8
--- a/Source/GUI/VCL/GUI_Main.h
+++ b/Source/GUI/VCL/GUI_Main.h
@@ -243,7 +243,6 @@ __published:    // IDE-managed Components
     void __fastcall M_Debug_HeaderClick(TObject *Sender);
     void __fastcall M_File_ExitClick(TObject *Sender);
     void __fastcall M_File_Open_FileClick(TObject *Sender);
-    void __fastcall FormShow(TObject *Sender);
     void __fastcall M_Debug_AdvancedClick(TObject *Sender);
     void __fastcall M_Options_ShowToolBarClick(TObject *Sender);
     void __fastcall M_View_EasyClick(TObject *Sender);
--- a/Source/GUI/VCL/GUI_Preferences.cpp
+++ b/Source/GUI/VCL/GUI_Preferences.cpp
@@ -389,6 +389,18 @@ void __fastcall TPreferencesF::CB_InfoTipClick(TObject *Sender)
 }
 
 //---------------------------------------------------------------------------
+void __fastcall TPreferencesF::CB_RememberWindowPositionClick(TObject *Sender)
+{
+	Prefs->Config(__T("RememberWindowPosition"), 1) = CB_RememberWindowPosition->Checked ? __T("1") : __T("0");
+}
+
+//---------------------------------------------------------------------------
+void __fastcall TPreferencesF::CB_RememberWindowDimensionsClick(TObject *Sender)
+{
+	Prefs->Config(__T("RememberWindowDimensions"), 1) = CB_RememberWindowDimensions->Checked ? __T("1") : __T("0");
+}
+
+//---------------------------------------------------------------------------
 void __fastcall TPreferencesF::CB_ShowToolBarClick(TObject *Sender)
 {
     Prefs->Config(__T("ShowToolBar"), 1)=CB_ShowToolBar->Checked?__T("1"):__T("0");
@@ -581,7 +593,7 @@ void __fastcall TPreferencesF::TreeChange(TObject *Sender,
 //---------------------------------------------------------------------------
 void __fastcall TPreferencesF::General_Language_MoreClick(TObject *Sender)
 {
-    Tree->Select(Tree->Items->Item[7]);
+    Page->ActivePage = Customize_Language;
 }
 
 //---------------------------------------------------------------------------
@@ -602,17 +614,16 @@ void __fastcall TPreferencesF::General_Output_MoreClick(TObject *Sender)
 //---------------------------------------------------------------------------
 void __fastcall TPreferencesF::FormShow(TObject *Sender)
 {
+    // review: GUI_Configure should be in FormCreate
     GUI_Configure();
     General_Output_SelChange(NULL);
 
-    //Not done with BCB because I want to easy select tabs in it
-    //Page->Top=-6;
-    //Page->TabHeight=1;
-    Page->Top=-(Page->TabHeight*1.15);                //Replaced above with this to hide tabs better on high-DPI
-    Page->Height=Page->Height+(Page->TabHeight*1.15); //Required with above line
-    Cancel->Top=Page->Top+Page->Height+(Cancel->Height*0.15);
-    OK->Top=Cancel->Top;
-    ClientHeight=OK->Top+(OK->Height*1.15);
+    // review: move to GUI_Configure
+	for (int cTabIndex = 0; cTabIndex <= Page->PageCount-1; cTabIndex++) {
+		Page->Pages[cTabIndex]->TabVisible = False;
+	}
+
+    // ClientHeight=OK->Top+(OK->Height*1.15);
 }
 
 //---------------------------------------------------------------------------
@@ -628,6 +639,8 @@ void __fastcall TPreferencesF::Setup_GeneralShow(TObject *Sender)
 //---------------------------------------------------------------------------
 void __fastcall TPreferencesF::Setup_AdvancedShow(TObject *Sender)
 {
+	CB_RememberWindowPosition->Checked=Prefs->Config(__T("RememberWindowPosition")).To_int32s();
+	CB_RememberWindowDimensions->Checked=Prefs->Config(__T("RememberWindowDimensions")).To_int32s();
     CB_ShowToolBar->Checked=Prefs->Config(__T("ShowToolBar")).To_int32s();
     CB_ShowMenu->Checked=Prefs->Config(__T("ShowMenu")).To_int32s();
     Advanced_CloseAllAuto->Checked=Prefs->Config(__T("CloseAllAuto")).To_int32s();
--- a/Source/GUI/VCL/GUI_Preferences.dfm
+++ b/Source/GUI/VCL/GUI_Preferences.dfm
@@ -1,12 +1,12 @@
 object PreferencesF: TPreferencesF
   Left = 286
   Top = 180
-  BorderIcons = []
-  BorderStyle = bsDialog
+  BorderIcons = [biSystemMenu, biMinimize]
   Caption = 'Preferences'
-  ClientHeight = 194
-  ClientWidth = 667
-  Color = clBtnFace
+  ClientHeight = 279
+  ClientWidth = 833
+  Color = clWindow
+  Constraints.MinHeight = 275
   Font.Charset = DEFAULT_CHARSET
   Font.Color = clWindowText
   Font.Height = -11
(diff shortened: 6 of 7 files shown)


JeromeMartinez:
[build with this PR](https://mediaarea.net/download/snapshots/binary/mediainfo-gui/20240702-2/MediaInfo_GUI_24.06.20240702_Windows.exe).
@regs01 thank you for your proposal, it is very appreciated.
Unfortunately I face an issue when I open the preference windows, I have an empty display: 
image
It is fine after I click on a page.
We use C++Builder 11 Community Edition.

@cjee21 is the change proposal about how to compute window sizes is fine at high DPIs?

PS: @regs01 please rebase, I just merged another PR and now it is conflicting, sorry about that, it is rare.

cjee21:
> @cjee21 is the change proposal about how to compute window sizes is fine at high DPIs?

I'm new to VCL so I just work on improving the existing codes. If there is a better and more proper way from someone more experienced in VCL then we should use that. I will test it later. We need to test on Windows 7 (to ensure no issues like previously) too and ideally it should be tested by someone with a multi-monitor, multi-DPI system. The existing implementation was not tested on this kind of set-up either so I don't know how it handles that.

It seems now the default is to not remember window position and remember window size. We need to consider whether to keep old behaviour as default (set both defaults to not remember) or not.

Also remember to use spaces for indents (it looks like tabs now).

JeromeMartinez:
> It seems now the default is to not remember window position and remember window size. We need to consider whether to keep old behaviour as default (set both defaults to not remember) or not.

I think it is a good intermediate.

> Also remember to use spaces for indents (it looks like tabs now).

True. @regs01 please use spaces instead of tabs.

cjee21:
> Unfortunately I face an issue when I open the preference windows, I have an empty display

Yes it's blank unless click on one of the items in the tree. Also it is entirely white with no distinction between the left, right and bottom sections. I don't know if that's a good idea. Also is it intended that the main window is re-spawned every time Preferences is closed?

The issue reported by Klaus1189 is still there and now is triggered as soon as Preferences is opened. Seems to take very long to open Preferences too.

cjee21:
> @cjee21 is the change proposal about how to compute window sizes is fine at high DPIs?

With remember window size disabled it is acceptable for me on different DPIs but it is different size/ratio than what MediaInfo has been using for years so I don't know how others will react.

With remember window size enabled, if user changes DPI after using MediaInfo at least once, it uses size for old DPI and window is too small (or maybe large). This may also cause issues for people using multi-monitors with different DPIs (I don't know how the behaviour is here since I no longer have multi-monitors).

JeromeMartinez:
>  it is different size/ratio than what MediaInfo has been using for years so I don't know how others will react.

I am not shocked by the default and not so different so it is acceptable for me.

>  if user changes DPI after using MediaInfo at least once, it uses size for old DPI and window is too small (or maybe large).

@regs01 is it possible to save a kind of DPI independent ratio rather than a pixel size?

cjee21:
> I am not shocked by the default and not so different so it is acceptable for me.

How default looks here:
Screenshot 2024-07-02 163338

Acceptable for me but no idea about other more particular users.

JeromeMartinez:
@cjee21 yours is not near from former version, ouch. too much height.

I have that:
image

Difference due to DPI somewhere?

JeromeMartinez:
@regs01 in practice we have 2 topics, the save window pos/dim on one side and the default size on another side. IMO it would be better to split the PR in 2: the first one keeps the same method for computing the default size. when it is merged, you send a PR with the changes you suggest about the default size.

cjee21:
> Difference due to DPI somewhere?

Not DPI. that's the default on 100% (96DPI). It is because I'm using higher resolution display with more screen estate. Not sure if it'll be even longer at 4K. So if there is a particular user who does not like to remember window size, that user may not be happy with a big change in the percentage of screen estate that MediaInfo uses by default. It is useful for text/HTML view but if the user uses Basic most of the time, that user may see that as wasted white space.

cjee21:
@JeromeMartinez Another major bug. When invoked from explorer context menu produces a window with no file open.

This change is causing lots of effects besides remember window size and change Preferences look.
- Preferences taking long time to open (I have a fast PC, what will happen on older PCs?)
- Main window re-spawning and flashing after Preferences closed (take longer and looks glitchy compared to the re-spawning we get when changing themes)
- The open via command line / context menu issue

cjee21:
At least it does not seem to break Windows 7 compatibility.

Note: We can remove delay load from user32 if we no longer need to get DPI manually.

cjee21:
We should set a minimum size if we allow resizing Preferences window:
Screenshot 2024-07-02 173200

cjee21:
Took a more detailed look at the changes in this PR.

I agree with changing the window size code to use VCL (`Monitor` and `ScaleFactor`) instead of the current one and remove all the native Windows API calls along with delay load of `user32`.

I also like that the window size/position saving is planned to be implemented as an option and that window position is disabled by default.

I think this should not be added:
```C++
    if (fpUnscaledMonitorHeight>=1440)
        Height=900;
```
as it is deemed "too much height" and may cause other users to complain although I can tolerate that
> @cjee21 yours is not near from former version, ouch. too much height.

This should not be simply removed from FormShow without sufficient testing and other changes:
```C++
    //Refresh global
    FormResize(NULL);
    Refresh();
```
These two functions have been triggered by FormShow long before I started contributing. There is a reason why it is there.

@JeromeMartinez these are my opinions since you asked me for a review. As for the other changes, I am not so sure but think they require more testing and changes due to many side-effects/issues mentioned earlier. Of course, you make the decision on what is to be done (I'm just a user/contributor) :)

JeromeMartinez:
> @JeromeMartinez these are my opinions since you asked me for a review.

Thanks :).


> As for the other changes, I am not so sure but think they require more testing and changes due to many side-effects/issues mentioned earlier. 

I guess that the best method here is to split this PR is 2 or more PRs with specific changes rather than all in one, too much difficult to review impacts.

cjee21:
Maybe it is a good idea to make both window size and position saving disabled by default so that users used to existing behaviour will not be affected and especially with the DPI change causing unintended window size. That is why I try to keep the window size the same as before even on high-DPI displays after adding high-DPI support. Then users who want their own size can turn the setting on and resize the window to their liking and deal with resizing it again when changing DPI since the change will be documented in changelog. If concerned about people not reading changelog, maybe can make MediaInfo display changelog on first run after upgrade like what some apps do.

regs01:
> Unfortunately I face an issue when I open the preference windows, I have an empty display:

I moved switching tab visibility before selecting active page.

> @cjee21 is the change proposal about how to compute window sizes is fine at high DPIs?

I commented in the line computing Preferences form height. There is no need as VCL scales it automatically. It's enough to have fine looking height in design time. I see also new commits to chage lables height for Language and Output. There is also no need for that. They just needed to be aligned to baseline of comboboxes, as they were aligned to bottom of comboboxes. VCL take control of scaling.

> We use C++Builder 11 Community Edition.

Yeah, I used C++ Builder 11.3 Community Edition. On VirtualBox, so it wouldn't mess up with old lifetime Delphi 10.2 Starter license. Although i'm using Lazarus now. But I must say how bad C++ Builder code completion is. It's so slow and buggy. Delphi's one is a little better. Even CnPack can't help much of situation. But Lazarus is completely different level - works instantly and only showing items within the scope. It just doesn't have prediction, as Visual Studio, but in everything else Lazarus code completion is even better than VS.

> True. @regs01 please use spaces instead of tabs.

I always use spaces, although RAD Studio seem to default tabs. There could be few, before I changed it. Probably CnPack also does it, i'll look up on it.

> Maybe it is a good idea to make both window size and position saving disabled by default so that users used to existing behaviour will not be affected and especially with the DPI change causing unintended window size. That is why I try to keep the window size the same as before even on high-DPI displays after adding high-DPI support. Then users who want their own size can turn the setting on and resize the window to their liking and deal with resizing it again when changing DPI since the feature will be documented in changelog. If concerned about people not reading changelog and knowing the existence of the feature, maybe can make MediaInfo display changelog on first run after upgrade like what some apps do.

I have 4K with 150% myself. It's just uncommon if users want to change scale factor on a daily basis. It is also highly unlikely that a person with high DPI display would use Windows 7. I left Position off by default, as this could be matter of preference. But with dimensions users wouldn't notice difference, as they will keep having default size and location. But those using Text, Tree views and similar will be able to quickly address form height issue without going into Settings.

> @regs01 please rebase, I just merged another PR and now it is conflicting, sorry about that, it is rare.

I'll look up on rebase. 

> We should set a minimum size if we allow resizing Preferences window:

I did for Height, as i didn't do scroll box. I can do for Width. It's 1-second operation.

> as it is deemed "too much height" and may cause other users to complain although I can tolerate that

#203 
There is no too much on large screen, like I have. Especially when you use Text/Tree views. I think most large screen owners will appreciate this, although I can remove it, considering Dimensions are saved on by default. So those using Text/Tree wound just be able to resize themselves.

> This should not be simply removed from FormShow without sufficient testing and other changes:

Cut off little bit less. I'll readd it. Although as I see at FormResize now, there is no need in most of it. All controls just need to be properly aligned.

cjee21:
One more thing. If MediaInfo is closed while maximized, the next time it will start un-maximized but filling up the entire screen. This is not a problem for me and I have seen other apps behave like this but other users may find it annoying. I suggest if possible and not too difficult, make it remember if it is maximized or not and also the window dimensions saved should be for un-maximized so that user can restore their original un-maximized size.

cjee21:
> is it possible to save a kind of DPI independent ratio rather than a pixel size?

We can save the window size at 96DPI and scale it accordingly when we read/write the value. That should work. It may be rare for a user to change DPI but what if there are people who use laptop and dock it on a monitor with a different DPI every day?

JeromeMartinez:
[dev snapshot](https://mediaarea.net/download/snapshots/binary/mediainfo-gui/20240703-3/MediaInfo_GUI_24.06.20240703_Windows.exe).

A lot better but it is still slower, as if something is called too many times (the language files maybe?).
the "new version available" test is also there but I guess it is a sidecar bug due to dynamic update of the interface without the right settings.

he dark mode is fine but in light mode it is super flat, some separation lines disappeared:
image

This PR has too many topics in one shot, it would still be better to split it in several PRs, one per topic, e.g. a first PR could be a resizable pref window and only that, and we merge it after validation then do a next step e.g. save window preferences without touching the algorithm for the default size then touching the  algorithm for the default size.

cjee21:
> [dev snapshot](https://mediaarea.net/download/snapshots/binary/mediainfo-gui/20240703-3/MediaInfo_GUI_24.06.20240703_Windows.exe).

Confirmed that invoking from CLI/context menu is still broken as I expected earlier after a quick look at the codes.

@JeromeMartinez I don't feel like reviewing this anymore
1. The indents are not fixed as desired by you even though this is the easiest thing
2. The complexity of the commits has increased instead of becoming clearer as you asked for and it was merged instead of rebased?
3. >This PR has too many topics in one shot, it would still be better to split it in several PRs
    
    Topics that are actually in this PR:
    1. Changing native Windows APIs to VCL APIs for DPI
    2. Adding new default window height for >1440p
    3. Implementing remember window position/size and the option
    4. Redesigning look of Prefs window
    5. Unknown changes that broke CLI invocation and caused main window to refresh on Preferences closed
4. I said to not remove `FormResize` and `Refresh` from `FormShow` to prevent breaking things. It was re-added but now `FormShow` is removed? So still broken and high risk of other things breaking that are yet to be discovered.
5. I do not get why it is required to redesign Preferences window
   - The existing one is perfectly serving it's purpose
   - The new all-white design without separation does not match with the rest of MediaInfo or other Windows apps
   - I don't see any benefit of being able to resize it, it's just all white space
   - It is now slower to open even noticeable on a 5GHz 1-year-old CPU (will it take ages for users on old PCs?)
   - The fix we added recently for aligning the buttons' heights on high DPI as requested by Klaus1189 has been undone:
      Screenshot 2024-07-03 221812
6. All these make me can't help but question if it was sufficiently tested and if there is a clear understanding of how the existing codes work before attempting change.
7. I suggest getting Klaus1189 to check the Preferences window instead of me since he appears to care about this window way more than me and to a pixel level. I only see this window once every update so I don't really care what happens here as long as can be used to change settings when required.

regs01:
> A lot better but it is still slower, as if something is called too many times (the language files maybe?).

You mean, when switching tabs? That's because ComboBoxes are filled every time tab is switched. I didn't touch it. It's an old behavior. A better option would be to fill them once at form creation. And when a language is added.

JeromeMartinez:
> You mean, when switching tabs? That's because ComboBoxes are filled every time tab is switched.

With the General tab, the one with languages.
I remark that the behavior is slower than without this PR, so something changed in this PR which is not related to its purpose.
Again, it would be easier for all if we have separate PR, one small PR per purpose, so we can easily catch the piece of code which changes the behavior.

> It's an old behavior.

There is definitely bad old behaviors but it is not good to make that worse while adding good things.

Your work is really appreciated but we need to find a way to avoid regressions when we add features.

regs01:
> Confirmed that invoking from CLI/context menu is still broken as I expected earlier after a quick look at the codes.

What context menu and where? Any steps to reproduce?


> I said to not remove `FormResize` and `Refresh` from `FormShow` to prevent breaking things. It was re-added but now `FormShow` is removed? So still broken and high risk of other things breaking that are yet to be discovered.

Everything is moved to GUI_Configure, which is called by constructor, as it should be. FormResize event is there. Deprecating it would be a next move.
https://github.com/regs01/MediaInfo/tree/mainformrearrange


> The fix we added recently for aligning the buttons' heights on high DPI as requested by Klaus1189 has been undone:

I didn't see the issue in VirtualBox with 125%, so I removed it. Check on my main system. Now I can see it. ComboBox does not scale properly. LCL has a feature to lock two controls to central lines of each other, as well as to top and or bottom line. Unfortunately VCL doesn't have it. Guess Embarcadero have no wish to develop VCL further. I'll brought it back, so as added to AfterMonitorDpiChanged.


> * The new all-white design without separation does not match with the rest of MediaInfo or other Windows apps

Since Windows 7 dialogs generally using clWindow color. I just made it the way so misalignments wouldn't be noticeable. But as there are no TabSheet borders now and if you don't like clWindow, it's can be easily changed to clBrnFace for TPanel and TForm. However I might suggest something in between - clWindow for TPanel and clBtnFace for Form.


> There is definitely bad old behaviors but it is not good to make that worse while adding good things.
> 
> Your work is really appreciated but we need to find a way to avoid regressions when we add features.

I literally didn't touch it. TabSheet events are same.
[screen2gif-2024-07-03.webm](https://github.com/MediaArea/MediaInfo/assets/10468217/e8924105-7079-46bf-8925-86a0b95953b8)

cjee21:
> What context menu and where? Any steps to reproduce?

Right click file in Windows Explorer > MediaInfo or using CLI `MediaInfo.exe %pathtoafile%` must not produce a MediaInfo window that looks like no file is open. 

> Everything is moved to GUI_Configure, which is called by constructor, as it should be. FormResize event is there. Deprecating it would be a next move. https://github.com/regs01/MediaInfo/tree/mainformrearrange

This one is up to @JeromeMartinez but since MediaInfo is working well right now, I think it is very risky to make changes that may break things and that require lots of testing especially since there are other more important issues that time should be spent on (including MediaInfoLib stuff). Also I'm not sure if the plan to replace this VCL version is still on. 

> I didn't see the issue in VirtualBox with 125%, so I removed it. Check on my main system. Now I can see it. ComboBox does not scale properly. LCL has a feature to lock two controls to central lines of each other, as well as to top and or bottom line. Unfortunately VCL doesn't have it. Guess Embarcadero have no wish to develop VCL further. I'll brought it back, so as added to AfterMonitorDpiChanged.

Everything should be tested upto 200% DPI scale as there are users using 200% that reported issues previously. 

> I literally didn't touch it. TabSheet events are same. Delay is exactly the same. You see `ComboBox_Update` call `RefreshFilesList`, which is reading all CSV language files. Slower the drive is, the longer delay would be. It might be also possible that you are using just one CSV file for development. So the only way to fix it is to move `ComboBox_Update` outside of drawing - into form creation.

The slowness I noticed and mentioned is when opening the Preferences window it takes longer to appear and I'm using a 5GHz CPU + DDR5 + PCIe 4.0 SSD. I'm also comparing with release version, not development.

As for the other things relating to Preferences window design, I suggest to just ask Klaus1189 who originally reported design issues with it instead of asking me.

Lastly I agree with splitting the PRs. I suggest the first PR for replacing the native Windows API calls as that can be merged immediately after one more quick test. Second one probably for implementing option to save window size/position since there is a feature request for it on MediaInfo website.

regs01:
> The slowness I noticed and mentioned is when opening the Preferences window it takes longer to appear and I'm using a 5GHz CPU + DDR5 + PCIe 4.0 SSD. I'm also comparing with release version, not development.
> 
> As for the other things relating to Preferences window design, I suggest to just ask Klaus1189 who originally reported design issues with it instead of asking me.

I just made a video. It's totally identical. There is nothing in my commits that affect it. But I can fix it anyway by moving reading CSV language files into form creation, so it just fires once. As well as a new language is added.



> Lastly I agree with splitting the PRs. I suggest the first PR for replacing the native Windows API calls as that can be merged immediately after one more quick test. Second one probably for implementing option to save window size/position since there is a feature request for it on MediaInfo website.

We can discuss it all first

> Right click file in Windows Explorer > MediaInfo or using CLI `MediaInfo.exe %pathtoafile%` must not produce a MediaInfo window that looks like no file is open.

I moved it down after reading args. Speaking of `Refresh` procedure it's better to rename it to avoid ambiguity, as `TForm` has own Refresh as a descendant of `TWinControl`. Could be named something like `UpdatePageContent`.

regs01:
Moving `ComboBox_Update` to `FormCreate`. And it was already in `Language_NewClick` and `Language_DeleteClick`. Same to be done to other ComboBoxes.

Link, if embeded video doesn't work in Firefox
https://github.com/MediaArea/MediaInfo/assets/10468217/af133654-9bad-4754-bee2-85337265370d

[screen2gif-2024-07-04-1955.webm](https://github.com/MediaArea/MediaInfo/assets/10468217/af133654-9bad-4754-bee2-85337265370d)

JeromeMartinez:
[New build](https://mediaarea.net/download/snapshots/binary/mediainfo-gui/20240704/MediaInfo_GUI_24.06.20240704_Windows.exe).
This is still slow but as slow as in [a build without this PR](https://mediaarea.net/download/snapshots/binary/mediainfo-gui/20240703-2/MediaInfo_GUI_24.06.20240703_Windows.exe), I guess that a slowdown was introduced in a a previous PR so not anymore a topic here but still need investigation in a separate issue or PR (@cjee21 I think it is from a previous PR made by you, before v24.06 as v24.06 also has it).

> We can discuss it all first, as it saves time

For the moment the only regression I still see is the lack of lines (they are there before this PR) in the light theme, also visible in the video, and it is not expected that theses lines change in a PR about main window position and re-sizable preference windows.

regs01:
I still can't get it. What soreness are you talking about? Is it one in the video?
https://github.com/MediaArea/MediaInfo/assets/10468217/e8924105-7079-46bf-8925-86a0b95953b8

I didn't make a commit yet. But there is a video with partial fix. I'll finish later, when come home.
https://github.com/MediaArea/MediaInfo/assets/10468217/af133654-9bad-4754-bee2-85337265370d

cjee21:
> [New build](https://mediaarea.net/download/snapshots/binary/mediainfo-gui/20240704/MediaInfo_GUI_24.06.20240704_Windows.exe). This is still slow but as slow as in [a build without this PR](https://mediaarea.net/download/snapshots/binary/mediainfo-gui/20240703-2/MediaInfo_GUI_24.06.20240703_Windows.exe), I guess that a slowdown was introduced in a a previous PR so not anymore a topic here but still need investigation in a separate issue or PR (@cjee21 I think it is from a previous PR made by you, before v24.06 as v24.06 also has it).

If we're still talking about the Preferences window, here are the commits history: https://github.com/MediaArea/MediaInfo/commits/master/Source/GUI/VCL/GUI_Preferences.cpp
All commits I made are after 24.06 release.
As for the function that launches Preferences window in GUI_Main, the only change I made was to FormResize to fix the blank toolbar visible despite being disabled issue that was reported. This only executes after Preferences window is closed and is after 24.06 release too. (https://github.com/MediaArea/MediaInfo/blame/65794412e6888ff1f9f68a8069600f787f56a6f2/Source/GUI/VCL/GUI_Main.cpp#L1625)

Time taken to open Preferences window:
| | |
| - | - |
| 24.04 (last release before I contribute/32-bit): | 0.851s |
| 24.06.20240702 (this PR's first snapshot/32-bit): | 1.602s |
| 24.06.20240703 (latest master?/64-bit): | 0.751s |
| 24.06.20240704 (this PR's latest snapshot/64-bit): | 1.017s |

I don't see any statistically significant difference regarding Preferences window after my contributions as expected from the commits history. The slightly faster time might be due to 64-bit and considering that a new option was added by your team recently, the results look good.

I don't know what happened but 1st version of this PR was bad. Now it can be considered okay. Take note that the above test results are likely not accurate to tens of ms (based on 30fps video & only run once) so hundred or so ms difference can be ignored.

> I still can't get what slowness are you talking about? Is it one in the video? https://github.com/MediaArea/MediaInfo/

I'm not sure which slowness he is talking about but the one I talk about in all my comments is time taken for the Preferences window to show up which was bad in 1st build of this PR but can be considered okay now for me.

regs01:
> UPDATE2: This still not reverted to previous fix:

Cherry picked to this branch

regs01:
> This takes ~310ms on 64-bit and ~530ms on 32-bit to execute:

This cannot take half a second. But once again - this is about to be moved from OnShow to OnCreate, so will take 0 ms.

regs01:
I made a commit here, so you measure it.

However, moving `ComboBox_Update` into `TForm.OnCreate` from `TTabSheet.OnShow` will make form opening slow anyway. But it's still better than having visual lags and flickering.

xerta555:
Hi

I've been waiting for this feature for a very long time. Apparently, there has been no progress since then.

Can we find out what's holding things up?
This is a crucial feature.
Also, is it possible to enable automatic line breaks?

Thx in advance.

regs01:
There was no response back then, so switched to other projects. But need response on relevant questions (not that there are tabs, not spaces, as it can be fixed in a second). But i still maintain VM with CPPBuilder (as don't want to mess up old Delphi on my main system), so i can return to it on New Year holidays.

bokeron2020:
@JeromeMartinez, it is time to address the elephant in the room.

This Pull Request is the definitive proof that the lack of basic UX features in MediaInfo is not a technical limitation or a "priority" issue, but a deliberate choice to spite the user base.

The request for a resizable window that remembers its position has been a top-tier grievance since 2007. Here, a contributor has done the heavy lifting, provided the code, and solved a 17-year-old problem for you. Yet, you let it sit here.

Let’s be clear: by intentionally keeping the free GUI crippled and archaic—even when working patches are handed to you on a silver platter—you are simply holding user ergonomics for ransom. You are leveraging your functional monopoly to create a "frustration gap," forcing people toward your paid App Store versions by refusing to implement OS standards that have been universal for three decades.

Using the "Open Source" label while acting like a cynical proprietary vendor who ignores the community is transparent. Rejecting or ignoring a working fix for a 17-year-old bug isn't "managing a backlog"; it is strategic stagnation and extortion.

Stop pretending you don't have the resources. You have the code right here in front of you. The only thing missing is the professional integrity to stop treating your users with contempt.

cjee21:
This PR would have been merged earlier if the contributor was willing to have a PR only for the remember window size and position feature. Most of the unresolved issues blocking this PR were due to the other changes made together in this PR.

By the way @JeromeMartinez you can remove me as reviewer for this. I no longer use the VCL version or have C++ Builder installed.

cjee21:
The poster above doesn't even know what he's talking about. There is no paid app store version for Windows and the one in Windows Store is outdated and even more lacking than the VCL one.

If wanting to proof something, this PR is proof that VCL is a poor framework where even features like this need so much manual work and causes many side effects. C++Builder Community Edition seems to be basically abandoned by Embarcadero. There are no new updates for over a year despite 3 newer paid versions available and the current free version being so buggy. They do not even support Windows on ARM or properly integrate with Windows 11 design, resulting in user complain about lack of rounded corners in menus.

Maybe it's time to switch to a better framework that requires less work for basic things and is easier to maintain as well as supporting newer OS and architectures along with being guaranteed to remain supported for free. Other open souce apps are moving from wxWidgets to Qt for example.

JeromeMartinez:
@cjee21 this is a bit more complicated, both technical and also about how I plan to do in the future as a whole.
Migration to Qt is one part of that, still not happy by the byte size of the deliverable, one of the appreciated "features" of MediaInfo being its size but less and less relevant...

regs01:
> This PR would have been merged earlier if the contributor was willing to have a PR only for the remember window size and position feature. Most of the unresolved issues blocking this PR were due to the other changes made together in this PR.
> 
> By the way @JeromeMartinez you can remove me as reviewer for this. I no longer use the VCL version or have C++ Builder installed.

It's not possible with GitHub. It had to be completely reset and then cherry picked line by line in diff into different branches. So first it had to be finished. And it mostly was finished. Although then there wouldn't to be much of need in those multiple requests.

> If wanting to proof something, this PR is proof that VCL is a poor framework where even features like this need so much manual work and causes many side effects. C++Builder Community Edition seems to be basically abandoned by Embarcadero.

VCL is by far more advanced and simpler than anything Qt can offer. It's just lack of experience with it that leads to all problems. But Qt is far more complicated.

And what exactly and what is proving what? Reading from many files will be as equally slow in Qt.

cjee21:
It takes more lines of code to do the same thing in VCL than it does in Qt. Dark mode in VCL is lacking aero titlebar on Windows 7, mica titlebar on Windows 11 and window shadows on all Windows versions. It respawns entire window when toggling dark mode. The menu bar's drop-down menu has no rounded corners on Windows 11, there is a user complain for that. The preferences window item has to be manually aligned taking much more time to add a single item compared to other frameworks. The items in preferences window does not auto resize when the preferences window is resized. The integration with IE or Edge WebView2 still glitches out sometimes with blank display for HTML view. VCL requires a custom compiler that even Embarcadero seems to have trouble keeping updated, it is still using very old Clang that does not support ARM64, no CFG, cannot compile code in parallel. How's that far more advanced. Since you are experienced in VCL, maybe you can fix some of the above. I am not the right person to be doing VCL work.

One more thing to improve for MediaInfo VCL version that is not related to VCL vs Qt framework is that the VCL version GUI hangs with not responding when MediaInfo takes longer to parse files due to it being run on main UI thread.

There's also user requests for ability to customize the font used.

Let's not even talk about the buggy C++Builder IDE that has broken autocomplete and code lookup among other issues and annoyances or the quality of Embarcadero's documentation for the C++ APIs.

regs01:
> Dark mode in VCL is lacking aero titlebar on Windows 7

Windows 7 doesn't have dark mode. And it doesn't matter since Qt6 doesn't support Windows 7 any longer.

> It takes more lines of code to do the same thing in VCL than it does in Qt.

Absolutely not. Especially if using Pascal. But even with C++ still VCL is far more simpler. And Qt6 doesn't have proper high DPI scaling. They ditching proper DPI scaling in favor of pixel ratio scaling, similar to browsers, which is unacceptable for applications.

> The menu bar's drop-down menu has no rounded corners on Windows 11, there is a user complain for that.

Neither in Qt. It's up to Win32 style. This is 6.10.1 with Win32 style.
menu-test_2026-01-24_09-46-02

> There's also user requests for ability to customize the font used.

It's just a component property. It's equally simple in VLC/LCL/WF/Qt etc etc.

> does not support ARM64,

Well, yes. It doesn't support ARM64 as of yet. It's where Free Pascal/Lazarus/LCL beats them all, supporting x86, ARM, RISC-V, MIPS, LoongArch etc etc and any type of system - Windows, Linux, macOS, other BSDs etc etc - trully cross platform.

> One more thing to improve for MediaInfo VCL version that is not related to VCL vs Qt framework is that the VCL version GUI hangs with not responding when MediaInfo takes longer to parse files due to it being run on main UI thread.

RefreshFilesList is a part of MediaInfo itself, not MediaInfoGUI. It's not using VCL. And Qt version is not using RefreshFilesList. Also it doesn't read content of locale files. It just using filenames. That's why it happens fast. It also doing it once at form creation. That's why you don't see UI lag. It's, again, just matter of implementations. You can do the same with VCL. It's also doesn't store locales in files on disk, instead reading them from resource. Although it makes it a little slower.

Moving it to form creation was one of that change you didn't like.

> The integration with IE or Edge WebView2 still glitches out sometimes with blank display for HTML view.

There is no Edge in Windows 7. And VCL version is using TWebBrowser. Although there is TEdgeBrowser, at least in CppBuilder 12. But TWebBrowser/TCppWebBrowser are using SHDocVw.dll, which is Trident (Internet Explorer).

> The preferences window item has to be manually aligned taking much more time to add a single item compared to other frameworks. The items in preferences window does not auto resize when the preferences window is resized.

That could be fixed in a single click. VCL has better alignment capabilities than Qt. LCL is even more supperior, but no C++. And this was among fixes.

regs01:
I've reset my master and updated CppBuilder. In coming days i'll readd changes with multiple branches and multiple merge requests.

cjee21:
> Windows 7 doesn't have dark mode. And it doesn't matter since Qt6 doesn't support Windows 7 any longer.

Doesn't support (technically Windows 10 does not have dark mode for Win32 apps either) but users will still want dark mode and MediaInfo VCL does have dark mode on Windows 7 although it looks non-ideal. We can use older Qt to have a separate build for Windows 7 if needed but MediaInfo Qt is not yet deployed for Windows and will likely only be used to replace UWP one in Store so let's not talk about this.

> But even with C++ still VCL is far more simpler.

Simplify it then. Maybe I am not good in VCL and implemented dark mode in non-ideal way. In Qt it takes less than 10 lines of CPP to control dark mode and it works out-of-the-box on Windows 11 with zero lines of code if just following system settings. Well adding conformance button to VCL and Qt each took around 40 lines of code so fair enough but adding new setting still feels faster and easier in Qt, maybe due to MediaInfo's VCL codebase not in ideal condition.

> Qt6 doesn't have proper high DPI scaling.

Whatever they did, at least it works fine with zero additional code unlike VCL which requires so many bugfixes previously.

> This is 6.10.1 with Win32 style.

Mica titlebar, window shadows, rounded menu drop-downs...

Screenshot 2026-01-24 175503

> RefreshFilesList is a part of MediaInfo itself, not MediaInfoGUI. It's not using VCL. And Qt version is not using RefreshFilesList.

I am not taking about the preferences window and this is not related to framework as I said. It is just something to improve in VCL version. There should be a progress bar or something while MediaInfoLib is opening files instead of freezing the entire UI.

> Moving it to form creation was one of that change you didn't like.

You can move it now if it does not break anything. I am no longer touching the VCL codes anyway so do whatever you want.

> There is no Edge in Windows 7.

There is Edge WebView on Windows 7 and MediaInfo's VCL GUI currently will use Edge WebView 2 whenever available. It has also been tested on Windows 7. Any way, whether IE or Edge, the HTML view just feels fragile. Will be great if can be made more reliable.

> That could be fixed in a single click. VCL has better alignment capabilities than Qt.

Okay. Go align it so that there is no longer need to manually shift items manually pixel by pixel when adding a new option in the future.

> I've reset my master and updated CppBuilder. In coming days i'll readd changes with multiple branches and multiple merge requests.

Great. Just take note that the official build farm is still using C++Builder 11 to maintain compatibility with Windows 7 (according to Jerome, the binary built with ver 12 does not work on Windows 7 out-of-the-box) so it should still be build-able with version 11.

cjee21:
btw make sure all alignment is pixel-perfect on all DPI settings. If it is even off by 1 pixel, someone named Klaus1189 will create an issue report.

regs01:
> Okay. Go align it so that there is no longer need to manually shift items manually pixel by pixel when adding a new option in the future.

It was already among fixes.

> Great. Just take note that the official build farm is still using C++Builder 11 to maintain compatibility with Windows 7 (according to Jerome, the binary built with ver 12 does not work on Windows 7 out-of-the-box) so it should still be build-able with version 11.

Community edition has limited time license. Mine run out. It's not possible to obtain RAD Studio 11 licenses anymore. But it doesn't matter much.

> Whatever they did, at least it works fine with zero additional code unlike VCL which requires so many bugfixes previously.

But custom drawn components and icons are not crisp. Especially on non-integer scales. Because there is no direct pixel access.

> VCL GUI currently will use Edge WebView 2 whenever available

I don't see it in code. 

> In Qt it takes less than 10 lines of CPP to control dark mode and it works out-of-the-box on Windows 11 with zero lines of code if just following system settings. Well adding conformance button to VCL and Qt each took around 40 lines of code so fair enough but adding new setting still feels faster and easier in Qt, maybe due to MediaInfo's VCL codebase not in ideal condition.

But it won't be Win32 style. Essentially it comes down to matter of taste, but there is nothing wrong in prioritizing system style. I always use windowsvista theme with Qt apps. So it's still same lines of code to allow to select and switch it. 

Main menu of Qt version isn't nice, btw. At least light version. Just black outline instead of shadow. And way too rounded. And scaling has its bugs as well. Main menu isn't properly scaled. And tree view items aren't scaled.
MediaInfo_2026-01-24_14-46-47
MediaInfo_2026-01-24_14-43-13


> I am not taking about the preferences window and this is not related to framework as I said. It is just something to improve in VCL version. There should be a progress bar or something while MediaInfoLib is opening files instead of freezing the entire UI.

Well, Qt version has progress bar implemented, but it's not VCL fault, that it's not implemented in VCL version. But there isn't much need for progress bar for 1 second.

cjee21:
>> VCL GUI currently will use Edge WebView 2 whenever available

>I don't see it in code.

I thought you are very experienced with VCL. I wonder why you cannot see it.

> But custom drawn components and icons are not crisp. Especially on non-integer scales. Because there is no direct pixel access.

All of Microsoft's new apps and OS UI uses WinUI3 which can be considered 'custom drawn' like Qt. There are no sharpness issue with both. Older Win32 apps are the ones lacking crispness. Isn't custom drawing direct pixel access? But I don't want to talk about this anymore. Does not matter here. Actually VCL styles are doing custom drawing and it seems to be needed to get dark theme with it so what's the point.

> And way too rounded. And scaling has its bugs as well. Main menu isn't properly scaled. And tree view items aren't scaled.

It's not too rounded. This is Microsoft's standard menu for Windows. I do not see scalling issues in MediaInfo Qt. On 100% and 150%, everything looks exactly identical except the DPI unlike the other GUI where buttons and alignment are different. Traditional Win32 apps no longer look native on Windows 11 or meet the design guidelines. Qt actually attempts to follow the latest Microsoft design guidelines on Windows 11. Doing so requires custom drawing like Microsoft's own WinUI3 framework. There are issues with Qt's menu shadows and the roundedness does not fully match Windows style but a fix has been committed for Qt 6.11 which should use Windows provided border and shadows for the menus instead of manually drawing.


https://learn.microsoft.com/en-us/windows/apps/develop/ui/controls/images/menu-bar-submenu.png

MediaInfo Qt 6.11

Screenshot 2026-01-25 050717
Screenshot 2026-01-25 050020

> But there isn't much need for progress bar for 1 second.

Try opening 10+ 1GB MPEG-TS files and watch the GUI hang.

Also note that some users open thousands of files at once with MediaInfo GUI. There was such a user in one of the issues here that ran into some 32-bit bug.

And like I alredy said, this is not framework problem. Just that it is something that can be improved in VCL version to prevent UI hangs. Just letting you know since you may want to do something about it since you are good at VCL coding.

JeromeMartinez:
> One more thing to improve for MediaInfo VCL version that is not related to VCL vs Qt framework is that the VCL version GUI hangs with not responding when MediaInfo takes longer to parse files due to it being run on main UI thread. [...] Try opening 10+ 1GB MPEG-TS files and watch the GUI hang.

This is a problem with **my** code, I know, there is a lot to do on it, and at some point I'll refactor that. And this is a bit out of topic...

regs01:
> I thought you are very experienced with VCL. I wonder why you cannot see it.

I would be glad if you point me a place where that WebView2 declared and created. But so far I can only see TWebBrowser.

```
#include <SHDocVw.hpp>
class TWebF : public TForm
{
  __published:
    TWebBrowser *Browser;
    void __fastcall BrowserNavigateComplete2(TObject *ASender, IDispatch * const pDisp, const OleVariant &URL);
  private:
  public:
    __fastcall TWebF(TComponent* Owner);
    void __fastcall Execute();
};
```

> I do not see scalling issues in MediaInfo Qt.

I do. And I pointed on them with screenshots. "Kind of works for me" is not an answer.

> Try opening 10+ 1GB MPEG-TS files and watch the GUI hang.

I tried 100 GB files. Under 1 second per file. A 1000 could be long, but, again, threading is implemented in MediaInfo itself, not in Qt GUI. It can be used with VCL GUI as well. It is, however, requires some extensive rework to switch to Core unit. But wouldn't be long to implement some of VCL's threading solutions.

> Traditional Win32 apps no longer look native on Windows 11 or meet the design guidelines.

Win32 is Win32. WinUI is WinUI. They both natively looking. One is for desktop (what is nowadays some call business UI), another is adaptive UI. Qt UI is that desktop UI. Qt has QML for adaptive UI. You using either of them. It doesn't go together when parts of interface have giant oversized touch oriented elements and other small desktop and mouse oriented elements.

> There are no sharpness issue with both. 

But there is. Just because it's oversharing final image it doesn't means it sharp and DPI native.

> Isn't custom drawing direct pixel access?

No. With pixel ratio scaling pixel itself is scaled, so it's virtual. You can't draw a single pixel. At 200% minimum you can access is a group of 4 pixels. Basically saying it's raster scaling very same as System scaling. Just adds more sharpness and draws text after scaling.

cjee21:
> I would be glad if you point me a place where that WebView2 declared and created.

This is one thing that VCL does make easy.

https://github.com/MediaArea/MediaInfo/blob/8cf99f07a5df94479f797f8c641c610bd20fde46/Source/GUI/VCL/GUI_Main.dfm#L586

> I do. And I pointed on them with screenshots. "Kind of works for me" is not an answer.

I do not see any issues with your screenshot. The scalling issues I talk about are how using different scalling results in different layouts/looks/alignments. If you mean Qt's menu design is not spaced properly, that's not scalling issue but how they chose to design it. They aim to follow Microsoft's fluent design guidelines for the new Windows 11 style but are still tweaking it.

> Win32 is Win32. WinUI is WinUI.

But VCL's menu does not even look like Windows Device Manager's menu and that's why there is an issue created by someone here.

> No. With pixel ratio scaling pixel itself is scaled, so it's virtual.

I'll just let AI answer:

>The claim states that at 200% scaling, the minimum you can access is a group of 4 pixels. This is incorrect.
>
>Both Qt and WinUI 3 allow you to bypass "logical" coordinates and target the device-specific physical pixels if you really want to.
>
>    In Qt: You can use devicePixelRatio() to calculate the exact physical coordinates. If you use QPainter on a QImage or QPixmap with the correct device pixel ratio set, you can absolutely manipulate a single 1×1 physical pixel on a 4K display.
>
>    In WinUI 3: Through Composition surfaces or Win2D, you can render at the full native resolution of the screen. You aren't "locked out" of those extra pixels; the framework just abstracts them away by default to keep your UI from looking tiny.

and in normal use, the abstraction is helpful so that you do not need to manually calculate and scale everything. The underlying framework scales things before painting the raw pixels ensuring everything has the same look regardless of scale and remains sharp. They do custom drawing on the actual pixels not draw then scale by interpolation.

Win32 apps like Device Manager are the ones using GDI scalling and are not as sharp.

PS. btw frameworks like Qt, I mentioned that is requires less code which is true in some aspects. They have many built-in functions like handling temporary files or remembering window state and position. So they do not require 100 lines of code to manually do it. MediaInfo Qt already has remember window size and position implemented by @JeromeMartinez years ago but it took this long for VCL. By using such frameworks, more time can be spent working on other parts of MedaInfo. The buggy C++Builder IDE compared to Qt Creator and Visual Studio also means more time to write the same amount of code due to broken auto-complete and code look-up for example. The poor documentation by Embarcadero does not help too.

One thing VCL helps more than Qt is integration with IE and Edge, whereas Qt requires self-implementation of Edge WebView if not using QML at the moment unless wanting to bundle Qt WebEngine. VCL's integration with IE and Edge does not feel very solid though (glitches and crashes seen before) or maybe that's due to how MediaInfo GUI uses it.

Anyway, let's stop debating frameworks. We have said what we wanted and it's clear that we will not reach any conclusion.

Issue #913, opened 2024-07-13
[UI] [24.06] About dialog has text cut


superbonaci:
Some text is cut and other is very close to edges

![MediaInfo_24.06_About](https://github.com/user-attachments/assets/6021d256-405d-47e1-acba-68e20be9fef3)

superbonaci:
When installing the graph plugin there are some issue also.

JeromeMartinez:
> When installing the graph plugin there are some issue also.

Could you be more precise?

cjee21:
No text is cut but the close to edges issue should already be fixed by https://github.com/MediaArea/MediaInfo/commit/d5e8e94d3c3c67c68c041d1d3ab3ac5832eb29cf

superbonaci:
> > When installing the graph plugin there are some issue also.
> 
> Could you be more precise?

![graph](https://github.com/user-attachments/assets/33df64bb-ec77-40be-84a3-827960f4c157)

PR #916, opened 2024-07-15
Windows GUI: Fixes

cjee21:
- Fix layout of Plugin install window
  - Fix overlapping of buttons with text. Resolves [#913](https://github.com/MediaArea/MediaInfo/issues/913#issuecomment-2227054711).
- Use consistent letter case for ToolBar
  - Change `Toolbar` to `ToolBar` to be consistent.
- Fix issues found by static analysis
  - Fix the following issues found with [Cppcheck](https://github.com/danmar/cppcheck). Use `int` instead of `size_t` as [ItemIndex is of type `int`](https://docwiki.embarcadero.com/Libraries/Athens/en/Vcl.Controls.TCustomListControl.ItemIndex). `size_t` is unsigned and cannot be reliably compared with `-1`.
    ```
    Id: knownConditionTrueFalse
    CWE: 571
    Condition 'ItemIndex_Save!=-1' is always true

    Id: knownConditionTrueFalse
    CWE: 570
    Condition 'Position==-1' is always false
    ```
  - Fix the following issue found with [Cppcheck](https://github.com/danmar/cppcheck) by removing redundant check.
    ```
    Id: nullPointerRedundantCheck
    CWE: 476
    Either the condition 'Page_Sheet_X_Web[KindOfStream]' is redundant or there is possible null pointer dereference: Page_Sheet_X_Web[KindOfStream].
    ```
- Fix issues found by static analysis 2
  - Follow the following recommendations from [Cppcheck](https://github.com/danmar/cppcheck) for 6 occurrences
    ```
    Id: cstyleCast
    CWE: 398
    C-style pointer casting detected. C++ offers four different kinds of casts as replacements: static_cast, const_cast, dynamic_cast and reinterpret_cast. A C-style cast could evaluate to any of those automatically, thus it is considered safer if the programmer explicitly states which kind of cast is expected.
    ```
  - Follow the following recommendation from [Cppcheck](https://github.com/danmar/cppcheck) by renaming `I` to `i` for inside loop usage to prevent confusion since `I` is already used for `MediaInfoList` outside of loop
    ```
    Id: shadowVariable
    CWE: 398
    Local variable 'I' shadows outer variable
    ```
- Fix Sheet view buttons size on high-DPI
  - Fix 'Web' buttons on 'Sheet' view being too large and clipped off on high-DPI displays

- Remove unused string
  - Seems to have reduced the exe size by 0.5KB

### DIFF
--- a/Source/GUI/VCL/GUI_Main.cpp
+++ b/Source/GUI/VCL/GUI_Main.cpp
@@ -1077,7 +1077,7 @@ void __fastcall TMainF::Refresh(TTabSheet *Page)
                     else
                     {
                         Buffer[Count]=(int8u)'\0';
-                        Ztring Template=Ztring().From_UTF8((char*)Buffer);
+                        Ztring Template=Ztring().From_UTF8(reinterpret_cast<char*>(Buffer));
                         if (Template.FindAndReplace(__T("@SVG@"), S1)==0)
                             S1=__T("Invalid template");
                         else
@@ -1111,7 +1111,7 @@ void __fastcall TMainF::Refresh(TTabSheet *Page)
             F.Write(S1);
             F.Close();
             //Navigate
-            Page_Custom_HTML->Navigate((MediaInfoNameSpace::Char*)FileName_Temp.c_str());
+            Page_Custom_HTML->Navigate(const_cast<MediaInfoNameSpace::Char*>(FileName_Temp.c_str()));
             FormResize(NULL);
         }
         else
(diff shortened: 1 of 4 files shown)


JeromeMartinez:
@cjee21 please rebase this one.
(918 was merged before this one but it seems pretty independent so fine, and there is only a small merge conflict)

cjee21:
Should be okay now. I not yet test compile though.

cjee21:
Okay now.

JeromeMartinez:
[build](https://mediaarea.net/download/snapshots/binary/mediainfo-gui/20241007-4/MediaInfo_GUI_24.06.20241007_Windows.exe)

PR #918, opened 2024-07-30
Windows GUI: Upgrade to C++Builder 12 Community Edition

cjee21:
- Save project file with C++Builder 12
  - Clean-up and save project file using C++Builder 12
- Enable Win64x target platform
  - Enable the newer toolchain that is based on LLVM/Clang 15.0 [(comparison)](https://docwiki.embarcadero.com/RADStudio/Athens/en/Clang-enhanced_C%2B%2B_Compilers).
  - To build this version, use changes in https://github.com/MediaArea/ZenLib/pull/158 and build Win64x targets for zlib, ZenLib and MediaInfo by changing `Win64` to `Win64x` in [BuildRelease.bat](https://github.com/MediaArea/MediaArea-Utils/blob/80b55f1366085b99dcaf8152f323459e427c4451/build_release/BuildRelease.bat#L286).
- Various fixes and improvements
  - To make it compile and run successfully on the new C++Builder and Win64x target.

Note:
Looks like it is still not possible to enable CFG and CET even though the new bcc64x toolchain is based on LLVM/Clang 15.0.


JeromeMartinez:
We upgraded to BCB12 without the PRs, then used the PR with the legacy compiler, so merged. We'll test the Win64x build soon.

JeromeMartinez:
We [reverted the BCB12 upgrade](https://github.com/MediaArea/MediaArea-Utils/pull/302) and switched back to BCB11 because BCB12 compiled binary requires [Windows 10 Universal C Runtime (CRT)](https://www.microsoft.com/en-us/download/details.aspx?id=48234) on Win7 and it is that on a bare metal Win7:
unnamed

We keep the related PRs because it does not hurt to have them.

cjee21:
Now I see that the 64-bit versions have dependency on UCRT.

It ran when I tested with a clean install Windows 7 VM that day. Maybe somehow my Windows 7 ISO has UCRT built-in?

JeromeMartinez:
> Maybe somehow my Windows 7 ISO has UCRT built-in? Or it got installed by MSVC runtime?

I can not say about that, I can only see that with a 10 year old VM with Win7 SP1 installed 10 years ago, the behavior is different between BCB11 and BCB12 compilation (only that was changed), and I am afraid that there are still installs with a such configuration, without knowing how many people would be impacted.
As it currently doesn't bother me to stay with BCB11, I think I'll stay with that for the moment, and BCB12 + Win64x will be activated when we drop Win7 support for a big reason. Currently I still have 5% of my users with Win7 to Win8.
In the meantime our master branch has the code for it so you don't have to maintain a specific branch, and we'll accept patches (please do! If it is something you like to do) focused on BCB12 + BCC64X if they don't break BCB11 compilation.

Detection of the lack of UCRT and download+install of it could be an alternative, but some additional work (also managing cases without internet connexion on the target machine with a hint for downloading the missing stuff on another machine) and I don't see as needed until we have a good reason to use an up to date LLVM/Clang instead of a 8+ year old one enough for us for the moment, especially with BCC32X still using the old one.

Long story short, for the moment I prefer to stay as it is now (patches in source code for BCB12 compatibility but still released with BCB11).

cjee21:
Putting this here for future reference.

Known issue with BCC64X in version 12.1 Patch 1:
When switching to dark mode then back using options menu, the app crashes with a sound but no pop-up after appearing to successfully switch back to light. Starting the app in dark then switching to light via menu crashes too. Interestingly, switching from light to dark and back then repeating many times by switching Windows' settings while the app is running does not trigger a crash. This should be tested on newer versions in the future before migrating to BCC64X.

Expected updates to BCC64X in version 12.2 (no community edition available yet):
Parallel compilation support to speed up compilation of multiple source files using multiple CPU cores.
Parallel compilation is already supported by the other two compilers but is not currently enabled in the project file.

cjee21:
> We [reverted the BCB12 upgrade](https://github.com/MediaArea/MediaArea-Utils/pull/302) and switched back to BCB11 because BCB12 compiled binary requires [Windows 10 Universal C Runtime (CRT)](https://www.microsoft.com/en-us/download/details.aspx?id=48234) on Win7 

According to https://support.microsoft.com/en-us/topic/update-for-universal-c-runtime-in-windows-c0514201-7fe6-95a3-b0a5-287930f3560c, it is auto-installed by Windows Update on Windows 7 and above so there should be no issues.

> Detection of the lack of UCRT and download+install of it could be an alternative

There appears to be no UCRT redistributable since it is installed by Windows Update.

JeromeMartinez:
> There appears to be no UCRT redistributable since it is installed by Windows Update.

My Win7 VM is maybe not up to date (rarely on), so maybe the UCRT was not installed automatically. I am hesitating to launch the update for seeing as this old VM is also a good example of what I may face in practice with the spread of this version, I think it is not worth it to do that now (not enough advantage in practice with the new toolchain compared to the risk), we may let Win7 die a bit more (months? years? I don't know yet).

PR #919, opened 2024-08-01
Create id.csv

williamsp:
Add Indonesian (bahasa Indonesia) translation file


JeromeMartinez:
Thank you for your translation proposal.

I adapted a bit the result for coherence with other files, please review the changes and squash the commits in 1 commit (else I'll do it but you will lose the GitHub "verified" tag).

williamsp:
Hi, thank you for sorting the file and add untranslated strings to the file.
On the last commit, I already translate the strings you have added before.
Please review, and update the carriage return, ZZ_Automatic_Percent, and ZZ_AutomaticLanguage_Percent if necessary.
Thank you.

JeromeMartinez:
I squashed all in one commit, FYI if you target 100% (not mandatory) remaining lines are:
ComplementaryObject
HomeTheaterAvr
Lines_MaxCountPerEvent
Position_Cartesian
Position_Polar
RealtimeLoudnessCorrected

Pull then force push (all in one commit in case you update the patch) if you wish to add the "verified" flag to your commit on GitHub.

williamsp:
Ah, I missed those lines. But, all of them already translated now.
For the squash and force push, I cannot find them in Github Web, so I try it from console instead. But I'm not sure if I try the right command. Please advise. Thank you.

JeromeMartinez:
> For the squash and force push, I cannot find them in Github Web, so I try it from console instead. But I'm not sure if I try the right command. Please advise. Thank you.

I don't use the "verified" stuff, here I see that the commit is unique and you are the only one assigned (when I pushed it was with both you and me) but still lacking the "verified" flag which was present in the past. Not a big deal.
I'll add another commit with the automatic stuff then merge.

williamsp:
Ah, I see. I think it's fine if just the verified flag that is missing.

PR #927, opened 2024-09-23
Fixes for Belarusian translation

zubialevich:
Hi,

Here are some fixes for existing Belarusian translation and also translation of some not translated lines


JeromeMartinez:
I updated `be.csv` with it and also added your name and email address in the list of authors with the ones of this patch, let me know if you prefer to keep them only in the patch.

zubialevich:
Thank you @JeromeMartinez !

PR #928, opened 2024-09-26
Update eu.csv

alexgabi:
Update basque language (eu)


Source/Resource/Plugin/Language/eu.csv
```diff
@@ -1,6 +1,6 @@
   Language_ISO639;eu
-  Author_Email;azpidatziak@gmail.com
-  Author_Name;Xabier Aramendi
+  Author_Email;alexgabi@disroot.org
```
JeromeMartinez, line 2: Put the older name in `Author_OldNames` field.

JeromeMartinez (changes requested):

PR #931, opened 2024-10-04
New MediaInfo.ico for consistency

cjee21:
This makes the `MediaInfo.exe` icon match the one used on the About window, MediaInfo website and elsewhere.


JeromeMartinez (changes requested):
Good catch.
No need to move to "old", it will be in the git history, please remove it from the PR.

JeromeMartinez (approved):

PR #944, opened 2024-11-08
Windows GUI: Improve start-up initialization

### LINKED ISSUE #942: Windows 11 V24.11 Opens with blank white screen
Every time I invoke it (right click), it display a blank white screen. I have to Select 'View' for it to to display the info. Went back to 24.06; works fine.

### LINKED ISSUE #943: Access Violation at adress
Sans titre
Windows 10 x64 V24.11
every time i launch a file identification, info are good but this message appear
back to 24.06 and ok

cjee21:
Ensure correct order during start-up to prevent bugs. Also reverts d0c7c2c1b7c3ec3c85e47111598fbc8de4897d71.

Fixes #942 
Fixes #943

### DIFF
--- a/Source/GUI/VCL/GUI_Main.cpp
+++ b/Source/GUI/VCL/GUI_Main.cpp
@@ -145,6 +145,7 @@ void __fastcall TMainF::ConfigTheme()
             }
             break;
     }
+    Refresh();
 }
 //---------------------------------------------------------------------------
 
@@ -262,19 +263,6 @@ __fastcall TMainF::TMainF(TComponent* Owner)
     if (I == NULL)
         I = new MediaInfoList;
 
-    //Load GUI preferences
-    GUI_Configure();
-
-    //Set Edge WebView2 UDF directory environment variable
-    Ztring UserDataDir=Prefs->BaseFolder;
-    UserDataDir.resize(UserDataDir.size()-1);
-    UserDataDir=UserDataDir.substr(0, UserDataDir.rfind(__T("\\"))+1);
-    UserDataDir+=__T("WebView2");
-    SetEnvironmentVariable(__T("WEBVIEW2_USER_DATA_FOLDER"), UserDataDir.c_str());
-
-    // Intitialize views for HTML output (prevent graphviz plugin from reading invalid metrics)
-    Page_Custom_HTML->Navigate(L"about:blank");
-
     //File(s) in command line
     #ifdef UNICODE
         if (IsWin9X())
@@ -293,6 +281,9 @@ __fastcall TMainF::TMainF(TComponent* Owner)
         for (int I1 = 1; I1 < ParamCount(); I1++)
             I->Open(Ztring().From_Local(ParamStr(I1).c_str()));
     #endif
+
+    //Load GUI preferences
+    GUI_Configure();
 }
 
 //***************************************************************************
@@ -310,8 +301,6 @@ void __fastcall TMainF::GUI_Configure()
         PreferencesF->Cancel->Enabled=false;
         PreferencesF->ShowModal();
         delete PreferencesF;
-        Prefs->Config_Load(); //Again...
-        #endif //MEDIAINFOGUI_PREFS_NO
 
         //Quick temporary fix for translations after an update
         //----------------------------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


JeromeMartinez:
[tested](https://mediaarea.net/download/snapshots/binary/mediainfo-gui/20241108-2/MediaInfo_GUI_24.11.20241108_Windows.exe), but error at start up then quit:
image

cjee21:
> but error at start up then quit

🤔 Just downloaded the installer from the link and performed a clean install (purging appdata). No issues but did find a bug where if it starts up with dark mode, window position is wrong.

cjee21:
> but error at start up then quit

🤔 Just downloaded the installer from the link and performed a clean install (uninstall and delete appdata before install). No issues. I tested the 32-bit version by extracting from the installer as well just in case, that one runs fine too.

Screenshot 2024-11-08 204050

I did find a bug where if it starts up with dark mode, window position is wrong (should be fixed with force-push above).

JeromeMartinez:
> Reverting [commit d0c7c2c](https://github.com/MediaArea/MediaInfo/commit/d0c7c2c1b7c3ec3c85e47111598fbc8de4897d71) seems to fix this. What is that commit supposed to fix?

Issue with the graph plugin and size of graphs.
Less important than crashes, please push a commit reverting that, I launch a build with all of that and we see on user machines.

g-maxime:
> Screenshot 2024-11-08 204654
> 
> Okay... some issue with IE engine that happens when WebView2 is not available.
> 
> Reverting [commit d0c7c2c](https://github.com/MediaArea/MediaInfo/commit/d0c7c2c1b7c3ec3c85e47111598fbc8de4897d71) seems to fix this. What is that commit supposed to fix?

This is supposed to fix another "division by zero" error dialog...

I'll check for an alternate fix (if still required with cairo)

JeromeMartinez:
@cjee21 please push the revert and I launch a build.

cjee21:
https://mediaarea.net/download/snapshots/binary/mediainfo-gui/20241108-3/MediaInfo_GUI_24.11.20241108_Windows_x64.exe

This build seems okay on Windows 11 with WebView2 installed except that in the condition where MediaInfo is set to default to both dark mode and HTML view, there is "Unspecified Error" and it fails to run.

JeromeMartinez:
[new build](https://mediaarea.net/download/snapshots/binary/mediainfo-gui/20241108-3/MediaInfo_GUI_24.11.20241108_Windows.exe).
Still same crash. Reverted to 24.11 and no crash.

cjee21:
Some more re-ordering. See if this works.

cjee21:
Seems there are two separate crashes here.

- Crash on startup with "Unspecified Error"
  - Condition: Set to default to HTML or Graph view + start in dark mode + Edge WebView2 runtime installed
  - Suspected cause: Interaction between VCL style switching and webview
- Crash on exit with "Access violation..."
  - Condition: Close window on system without Edge WebView2 runtime
  - Suspected cause: Interaction between commit https://github.com/MediaArea/MediaInfo/commit/d0c7c2c1b7c3ec3c85e47111598fbc8de4897d71 and webview when using IE engine

cjee21:
https://mediaarea.net/download/snapshots/binary/mediainfo-gui/20241108-4/MediaInfo_GUI_24.11.20241108_Windows_x64.exe

Seems above crashes and the original issue that this PR is supposed to fix is no longer reproducible.

There is an old bug remaining that only happens in this specific condition though:
WebView2 not installed -> Open MediaInfo -> switch to HTML -> switch to basic -> change theme -> change to HTML -> "Unspecified error' and HTML view cannot be used until MediaInfo restart

JeromeMartinez:
> Seems above crashes and the original issue that this PR is supposed to fix is no longer reproducible.

No more crash on my side, and [confirmed to work](https://sourceforge.net/p/mediainfo/bugs/1196/#c942/2624) by another person.

> There is an old bug remaining that only happens in this specific condition though:

Would enjoy that you fix it at some point but I am fine if too complicated and that stays.

JeromeMartinez:
Confirmed to work also with people having sent an email.

JeromeMartinez:
> By the way, will the Android version be updated on Google Play? Google Play is still offering 24.05.

It takes always more time for stores, but it is on the way.

JeromeMartinez:
@cjee21 see  https://github.com/MediaArea/MediaInfo/issues/945#issuecomment-2465408069 + I got a report ctrl-a ctrl-c doesn't work anymore.
-------- Message d'origine --------De : cjee21 ***@***.***> Date : 08/11/2024  17:42  (GMT+01:00) À : MediaArea/MediaInfo ***@***.***> Cc : Jérôme Martinez ***@***.***>, State change ***@***.***> Objet : Re: [MediaArea/MediaInfo] Windows GUI: Improve start-up initialization (PR #944) 

Would enjoy that you fix it at some point but I am fine if too complicated and that stays.

Since I have no idea about the cause, it only happens under specific conditions that are unlikely in normal use and it only affects the old IE engine that will probably be retired and all Windows 11 PCs will have WebView2, I will not be looking into it.
By the way, will the Android version be updated on Google Play? Google Play is still offering 24.05.

cjee21:
> ctrl-a ctrl-c doesn't work anymore

CTRL+A does not work in text view
CTRL+C works after highlighting in text view

Maybe useful: https://stackoverflow.com/questions/8466747/automatically-allowing-ctrla-to-select-all-in-a-tmemo

cjee21:
> > By the way, will the Android version be updated on Google Play? Google Play is still offering 24.05.
> 
> It takes always more time for stores, but it is on the way.

It's taking so long to roll out. I just got the 24.06 update from Google Play right now. 24.11 should be rolled out more quickly because HTML is buggy in 24.06 with overlapping text.

PR #946, opened 2024-11-08
Text view changes for Windows and Qt GUI

cjee21:
- Make VCL GUI Text views editable again due to requests
- Make Qt GUI Text view editable to match VCL GUI behaviour
- Reduce brightness of Text and Tree views in VCL GUI light mode to exactly match the brightness of Easy view and also other Windows apps such as Notepad.

Fix https://github.com/MediaArea/MediaInfo/pull/944#issuecomment-2465416491

### DIFF
--- a/Source/GUI/Qt/mainwindow.cpp
+++ b/Source/GUI/Qt/mainwindow.cpp
@@ -514,15 +514,15 @@ void MainWindow::refreshDisplay() {
             default:
             case VIEW_TEXT:
                 C->Menu_View_Text();
-                viewWidget = new QTextBrowser();
-                ((QTextBrowser*)viewWidget)->setFont(font);
+                viewWidget = new QTextEdit();
+                ((QTextEdit*)viewWidget)->setFont(font);
                 if(ConfigTreeText::getIndex()==0)
-                    ((QTextBrowser*)viewWidget)->setText(wstring2QString(C->Inform_Get()));
+                    ((QTextEdit*)viewWidget)->setText(wstring2QString(C->Inform_Get()));
                 else {
                     for (size_t FilePos=0; FilePos<C->Count_Get(); FilePos++) {
                         for (int streamKind=0;streamKind<4;streamKind++) {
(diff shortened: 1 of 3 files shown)


JeromeMartinez:
Looks like that a mistake (having the text output editable) is actually a wanted feature, could you bring back the editable text view?

Received by email:

The latest update is useless when the edit / Select all not working!

Ctrl+A: Not working!
Edit: If I wanna delete the path folder can't do anymore! It was good before this update!

I hope you guys let us control back! Instead of copy paste into Notepad to do that!

JeromeMartinez:
> FYI the wxwidgets (tested on Ubuntu) and Qt (tested on Windows) ones are not editable.

There are 100x more users on Windows and then 100x more user feedback/complains on it :-p.
(could also be done on Wx & Qt but not a priority for me, priority is to avoid receiving too many emails from Windows users)

JeromeMartinez:
>  if they [...]

A rule for developers of widespread software: don't try to understand all of them, you will become crazy before you understand.

JeromeMartinez:
[snapshot](https://mediaarea.net/download/snapshots/binary/mediainfo-gui/20241109-2/MediaInfo_GUI_24.11.20241109_Windows.exe) (before brightness).

JeromeMartinez (approved):

cjee21:
@JeromeMartinez side note:

Actually I wanted to make MediaInfo dark mode look nicer with native title bar and window shadows like below:

Dark mode concept test 2024-11-10 165756

However some issues regarding VCL styles and `MainMenu` prevents this from being a success. If a different bar that is not `TMainMenu` is used for menu bar, it will be easy but migrating the menu bar seems too much work.

JeromeMartinez:
> it will be easy but migrating the menu bar seems too much work.

It is nice from you to try, but on our side we won't change that because we need to focus on format support and we don't want to refactor everything now (as you can see, any change in the UI creates a lot of feedback...).

The UI you created is already very nice (based on from where it comes from, without changing the basics) by having high DPI and Dark Theme as it is!

PR #952, opened 2024-11-13
Windows GUI: Reset Text and Tree views colours to default

cjee21:
The final change for Tree and Text view background colour. Use the default colour selected by VCL / Windows for these kind of views. Any dissatisfaction is no longer the responsibility of MediaInfo's codes.

### DIFF
--- a/Source/GUI/VCL/GUI_Main.dfm
+++ b/Source/GUI/VCL/GUI_Main.dfm
@@ -489,7 +489,6 @@ object MainF: TMainF
           Top = 121
           Width = 604
           Height = 104
-          Color = 16382457
           EditMargins.Auto = True
           Font.Charset = DEFAULT_CHARSET
           Font.Color = clWindowText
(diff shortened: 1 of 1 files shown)


JeromeMartinez:
Maybe checking the actual color and if > 249 we set 249? So it is not too much white when white is the default and we keep other default values.

cjee21:
> Maybe checking the actual color and if > 249 we set 249? So it is not too much white when white is the default and we keep other default values.

So far I have been setting colours using the GUI form editor and do not know/want to write codes for that. I do not think 255 is an issue because it is used by almost all Windows apps by default including web browsers and File Explorer. Only some apps use 249 but I do not see any 249 in the list of colours defined in Windows. If it is an issue, then it would not be defined as the default window colour of Windows. Those who have issues with it would have the same issues with many other apps and should use a different solution.

JeromeMartinez:
> So far I have been setting colours using the GUI form editor and do not know/want to write codes for that. 

I'll give it a try. Not keen of 255.

PR #953, opened 2024-11-13
Qt GUI: Use WebEngine for HTML view

cjee21:
Use WebEngine for HTML view to render the HTML properly.


cjee21:
PS if the VCL and Qt versions were to be officially released side by side for Windows, I would have picked the Qt one.

PR #954, opened 2024-11-14
Qt GUI: Fixes and improvements

cjee21:
Qt GUI: Make toolbar moveable but prevent it from being detached
- Correction of 43289a54085aa55d5a881d94f1278a942138501e
- Turns out the toolbar was designed to be moveable. The main intention of the previous commit was to prevent toolbar from being detached from main window. Now it is moveable but must be docked to one of the four sides of the window.

Qt GUI: Save shell extension preference
- Prepare for future integration with IExplorerCommand-based shell extension on Windows.
- The saved preference can be read from registry by the IExplorerCommand implementation to decide if it should be visible.

Qt GUI: Add inform version and timestamp options
- Something that the VCL GUI has

Qt GUI: Add captions display options
- Match other toolkit versions

### DIFF
--- a/Source/GUI/Qt/mainwindow.cpp
+++ b/Source/GUI/Qt/mainwindow.cpp
@@ -515,6 +515,8 @@ void MainWindow::refreshDisplay() {
             default:
             case VIEW_TEXT:
                 C->Menu_View_Text();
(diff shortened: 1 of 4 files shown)


cjee21:
Screenshot 2024-11-14 193329

I don't have a suitable file to test display captions so hope I didn't make mistake and it works.

JeromeMartinez (approved):

PR #955, opened 2024-11-14
Qt GUI: Fixes and improvements 2

cjee21:
Qt GUI: Revert 'Make Text view editable'
- Revert commit b9fbf9ec8aa9e15bebcc87eb4fe102f47fe56f5f as it broke drag and drop of files into the text view. Text view isn't supposed to be editable anyway so match the wxWidgets version instead of VCL one.

Qt GUI: Add legacy stream display option
- An option that is present in MediaInfoLib and CLI but not yet in the GUI
- Disabled by default

Screenshot 2024-11-14 204655

Qt GUI: Enable WithThread so that progress bar actually works
- There is progress bar implemented but it only flashes a split second after MediaInfo appears to hang while opening many files.
- This commit makes the progress bar serve its purpose

Progress bar

Qt GUI: Add FFmpeg plugin support
- Tested

Screenshot 2024-11-15 133957

Qt GUI: Add Graph view
- Qt GUI now has working graph view

| Light | Dark |
| - | - |
| Screenshot 2024-11-15 155806 | Screenshot 2024-11-15 155853 |

Qt GUI: Add Graph export
- Graph can be exported as SVG

### DIFF
--- a/Project/QMake/GUI/MediaInfoQt.pro
+++ b/Project/QMake/GUI/MediaInfoQt.pro
@@ -200,6 +200,7 @@ unix:LIBS += -ldl \
               -lz
 
 SOURCES += ../../../Source/GUI/Qt/main.cpp \
+    ../../../Source/GUI/Qt/graphplugin.cpp \
     ../../../Source/GUI/Qt/mainwindow.cpp \
     ../../../Source/Common/Core.cpp \
     ../../../Source/GUI/Qt/custom.cpp \
@@ -220,6 +221,7 @@ SOURCES += ../../../Source/GUI/Qt/main.cpp \
 HEADERS += ../../../Source/GUI/Qt/mainwindow.h \
     ../../../Source/Common/Core.h \
     ../../../Source/GUI/Qt/easyviewwidget.h \
+    ../../../Source/GUI/Qt/graphplugin.h \
     ../../../Source/GUI/Qt/prefs.h \
     ../../../Source/GUI/Qt/views.h \
     ../../../Source/GUI/Qt/custom.h \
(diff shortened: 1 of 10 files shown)


cjee21:
**Current state of Qt GUI after this PR:**

Better compared to VCL GUI:
- More native looking dark mode and seamless switching
- Option to remember window size, position and state
- No cropped text in easy view
- Text wrapping in text view
- CFG and CET enabled for entire GUI
- Legacy stream display option available
- Preferences window can be resized usefully (contents expand to make full use of the space)
- Faster compilation time
- Seems to be smoother with less graphics glitches
- Easy integration with modern Windows context menu

Lacking compared to VCL GUI:
- No graph view
- No FFmpeg plugin support
- Requires bundling Qt WebEngine on Windows
- No translations
- Shell InfoTip setting does nothing at the moment

JeromeMartinez:
It is great!
I am still reluctant for the full migration on Windows, especially for the HTML part (need to find something for that), but if you keep improving Qt I guess that I will at least drop the Wx version and put Qt version on all Linux platforms and also for the free version on macOS at some point...

> Easy integration with modern Windows context menu

How is it possible? Qt stuff? Not possible to use the same with the VCL version?

> No temporary file generated in %TEMP% for HTML and Graph view

Same, not something doable with VCL too?

JeromeMartinez (changes requested):
Please remove the build/* files

cjee21:
> Please remove the build/* files

Ooops, accidentally included. I thought I had checked files that were included.

JeromeMartinez (approved):

JeromeMartinez:
> By the way, should I implement the legacystreamdisplay option for the VCL version too?

Yes.
At some point the call to options will be though MediaInfoLib for avoiding duplicated code, but it is a good thing to have everything in sync before doing this work.

PR #959, opened 2024-11-16
Windows GUI: Add LegacyStreamDisplay option

cjee21:
Note: I did not implement translations for this preferences option.


JeromeMartinez:
@cjee21 FYI I had to recreate the dfm file in BCB11 because it keeps crashing with your version. [fix](https://github.com/MediaArea/MediaInfo/pull/966).
(and at the same time I remove the hint for reloading because it is not the only one and at long term I'll handle such issue differently)

cjee21:
Just noticed this option affects display of "HDR Format" too. Maybe the option in Preferences needs a rename.

Issue #577, opened 2021-11-17
Rounded pixel aspect ratio


StefanPlayPlay:
Hello, 

I've encountered a problem when trying to process videos with non-square pixel aspect ratios based on information given by mediainfo. For demonstration purposes I've created two videos.

[First video](https://user-images.githubusercontent.com/90192257/142037181-165b56b8-2c9c-46e8-be20-44d796e7b90d.mp4):
```
Width: 1080
Height: 1490
Pixel aspect ratio : 149:108
```
Mediainfo given pixel aspect ratio : `1.380`

[Second video](https://user-images.githubusercontent.com/90192257/142037487-56ec3aec-4891-454d-877e-c8aceddeeade.mp4):
```
Width: 1490
Height: 1080
Pixel aspect ratio: 108:149
```

Mediainfo given pixel aspect ratio : `0.725`

So when I try to remove the non squareness of pixels I do something like : 
```
if(pixelAspectRatio > 1) {
 newWidth = width * pixelAspectRatio // Gives 1490,4 in the first video
} else {
 newHeight = height / pixelAspectRatio // Gives 1489,655172413793103
}
```

The results are not as expected because of the rounded pixel aspect ratio : 1080 * (149/108) should give 1490. 

So my question is : Is there any way I could retrieve the actual pixel aspect ratio & not the rounded one ? 

Thanks !

PR #960, opened 2024-11-16
Windows GUI: Windows 11 File Explorer Context Menu

### LINKED ISSUE #671: Compatible with Win11 new context menu (suggestion)
Compatible with Windows11 new context menu

cjee21:
Integrate with Windows 11 File Explorer modern context menu by implementing required items:
- Shell extension DLL which implements IExplorerCommand interface
- Sparse package which grants package identity
- Package helper DLL used by installer to add and remove the sparse package
- Updated installer script to handle everything

New shell extension also has the benefit of opening multiple selected files in the same MediaInfo instance instead of in multiple windows.

Resolves #671 

---

**Explanation of approach chosen for this new context menu implementation**

- The sparse package is provisioned for all users by the installer during install and removed for all users during uninstall using a helper dll. This should ensure all users will have the context menu on login after install and a clean removal after uninstall.

- The 'Assets' folder and `resources.pri` file are needed because when the app runs with app identity, the assets referenced by the sparse package manifest is used for the app icon. Since it is a sparse package, it itself does not contain assets and instead references the assets that are in the declared external location. The `resources.pri` is needed for the 'unplated' icons which are needed to prevent the icon from having accent-coloured plating as the background.

- In order to minimize the size of the DLL while ensuring it runs on clean Windows installations without Microsoft Visual C++ Redistributable installed, `vcruntime` is statically linked while `ucrt` is dynamically linked. This results in something between using `/MT` and `/MD`. It is known as [Hybrid CRT](https://github.com/microsoft/WindowsAppSDK/blob/main/docs/Coding-Guidelines/HybridCRT.md) and is supported according to the CRT maintainer.

- `/PDBALTPATH:%_PDB%` is added to linker in release mode so that the PDB file path is not contained in the DLL. This ensures no path information leakage and reduces the size of DLL slightly while still enabling analysis that requires PDB files to be done, for example SizeBench.

- The shell extension is re-written from WRL to C++/WinRT. See 'Note' at https://learn.microsoft.com/en-us/cpp/cppcx/wrl/windows-runtime-cpp-template-library-wrl?view=msvc-170 for details on why.

**References**
- https://blogs.windows.com/windowsdeveloper/2021/07/19/extending-the-context-menu-and-share-dialog-in-windows-11/
- https://blogs.windows.com/windowsdeveloper/2019/10/29/identity-registration-and-activation-of-non-packaged-win32-apps/
- https://learn.microsoft.com/en-us/windows/apps/desktop/modernize/grant-identity-to-nonpackaged-apps
- https://learn.microsoft.com/en-us/windows/win32/api/shobjidl_core/nn-shobjidl_core-iexplorercommand
- https://github.com/microsoft/AppModelSamples/tree/master/Samples/SparsePackages
- https://github.com/microsoft/vscode-explorer-command
- https://github.com/xandfis/W11ContextMenuDemo
- https://github.com/M2Team/NanaZip/blob/main/NanaZip.UI.Modern/NanaZip.ShellExtension.cpp
- https://github.com/notepad-plus-plus/notepad-plus-plus/blob/ce4d374a4782185f2e0dbaa2e6527ba7f8b9ad39/PowerEditor/src/tools/NppModernShell/Installer.cpp
- https://github.com/notepad-plus-plus/nppShell/blob/5e410dd6450a2572dd2eb209250561b6eec97535/Installer.cpp
- https://learn.microsoft.com/en-us/windows/uwp/cpp-and-winrt-apis/intro-to-using-cpp-with-winrt
- https://learn.microsoft.com/en-us/cpp/cppcx/wrl/windows-runtime-cpp-template-library-wrl?view=msvc-170
- https://github.com/microsoft/wil

---
Todo for MediaInfo team:

- [x] Update version update script for:
  - `Source\GUI\VCL\Manifest.manifest`
  - `Source\WindowsSparsePackage\MSIX\AppxManifest.xml`
  - `Source\WindowsShellExtension\Resource.rc`
  - `Source\WindowsPackageHelper\Resource.rc`

- [x] Update build script
  - make sparse package
    `makeappx pack /d "Source\WindowsSparsePackage\MSIX" /p "Project\MSVC2022\x64\Release\MediaInfo_SparsePackage.msix" /nv`

  - build DLLs
     ```
     MSBuild /t:MediaInfo_WindowsShellExtension /restore /p:RestorePackagesConfig=true;Configuration=Release;Platform=x64 Project\MSVC2022\MediaInfo.sln
     MSBuild /t:MediaInfo_PackageHelper /restore /p:RestorePackagesConfig=true;Configuration=Release;Platform=Win32 Project\MSVC2022\MediaInfo.sln
     ```

  - sign sparse package MSIX and DLLs generated by above

- [ ] Test on other system configurations besides Windows 11 Home single account.


cjee21:
Note: Open source code licensed under GPL and MIT from Microsoft, NanaZip and Npp were referenced in the making of this PR. Although code in this PR is not identical to any of the referenced code, it should be ensured that there are no licensing issues. I am not a legal expert.

Source/Common/Preferences.cpp
```diff
@@ -264,6 +262,19 @@ int Preferences::Config_Save()
...
+    DWORD dwBuild=0;
+    if (dwVersion<0x80000000) dwBuild=(DWORD)(HIWORD(dwVersion));
+    if (dwMajorVersion==10 && dwBuild>=22000) {
+        Config(__T("ShellExtension"))=__T("0");
+        Config(__T("ShellExtension_Folder"))=__T("0");
+    } else {
+        if (Config(__T("ShellExtension")).empty()) Config(__T("ShellExtension"))=__T("1");
+        if (Config(__T("ShellExtension_Folder")).empty()) Config(__T("ShellExtension_Folder"))=__T("0");
+    }
+
```
JeromeMartinez, line 277: I don't see the added value of not using `ShellExtension` on both Win10- and Win11+.
In my opinion this part should not be changed and other parts of the PR should be adapted to `ShellExtension` value.

Source/GUI/VCL/GUI_Preferences.cpp
```diff
@@ -342,6 +342,20 @@ void __fastcall TPreferencesF::Language_EditClick(TObject *Sender)
...
+                if (CB_InscrireShell->Checked) {
+                    Reg_HideExplorerContextMenu->DeleteValue("HideExplorerContextMenu");
+                }
+                else {
+                    Reg_HideExplorerContextMenu->WriteInteger("HideExplorerContextMenu", 1);
+                }
+            }
+        } catch (...) {}
+        delete Reg_HideExplorerContextMenu;
+    }
```
JeromeMartinez, line 358: See above (should not be needed).

Source/GUI/VCL/GUI_Preferences.cpp
```diff
@@ -360,6 +374,11 @@ void __fastcall TPreferencesF::CB_CheckUpdateClick(TObject *Sender)
 //---------------------------------------------------------------------------
 void __fastcall TPreferencesF::CB_InscrireShellClick(TObject *Sender)
 {
+    // Handle differently on Windows 11
```
JeromeMartinez, line 377: See above (should not be needed).

Source/GUI/VCL/GUI_Preferences.cpp
```diff
@@ -370,12 +389,16 @@ void __fastcall TPreferencesF::CB_InscrireShellClick(TObject *Sender)
 //---------------------------------------------------------------------------
 void __fastcall TPreferencesF::CB_InscrireShell_FolderClick(TObject *Sender)
 {
+    // Cannot be changed on Windows 11
```
JeromeMartinez, line 392: Isn't there a way in the DLL to see if it is a file or a folder and show the MediaInfo option depending on the setting?

Source/GUI/VCL/GUI_Preferences.cpp
```diff
@@ -624,8 +647,29 @@ void __fastcall TPreferencesF::Setup_GeneralShow(TObject *Sender)
 {
     ComboBox_Update(General_Language_Sel, Prefs_Language);
     CB_CheckUpdate->Checked=Prefs->Config(__T("CheckUpdate")).To_int32s();
-    CB_InscrireShell->Checked=Prefs->Config(__T("ShellExtension")).To_int32s(); //Lecture Shell extension
-    CB_InscrireShell_Folder->Checked=Prefs->Config(__T("ShellExtension_Folder")).To_int32s(); //Lecture Shell extension
+
+    if (TOSVersion::Major == 10 && TOSVersion::Build >= 22000) {
```
JeromeMartinez, line 651: See above (should not be needed).

Source/WindowsShellExtension/dllmain.cpp
```diff
@@ -0,0 +1,293 @@
+// dllmain.cpp : Defines the entry point for the DLL application.
+
+// IExplorerCommand File Explorer shell extension for context menu entry
+
+#include "pch.h"
+
+#ifdef MEDIAINFO_QT
+    #define SHELLEXT_GUID "dea3006d-451d-4f0f-a2b2-c3250cb255b4"
+#else
+    #define SHELLEXT_GUID "20669675-b281-4c4f-94fb-cb6fd3995545"
```
JeromeMartinez, line 10: Not possible to have 1 GUID for both?

Source/WindowsShellExtension/dllmain.cpp
```diff
@@ -0,0 +1,293 @@
...
+            hKey,
+            subKey.c_str(),
+            value.c_str(),
+            RRF_RT_REG_DWORD,
+            nullptr,
+            &data,
+            &dataSize
+        );
+        if (retCode != ERROR_SUCCESS)
+            throw std::runtime_error("Cannot read DWORD from registry.");
```
JeromeMartinez, line 92: I prefer silently quitting.

Source/WindowsShellExtension/dllmain.cpp
```diff
@@ -0,0 +1,293 @@
...
+        if (retCode != ERROR_SUCCESS)
+            throw std::runtime_error("Cannot read string from registry.");
+        DWORD stringLengthInWchars = dataSize / sizeof(wchar_t);
+        --stringLengthInWchars; // Exclude the NUL written by the Win32 API
+        data.resize(stringLengthInWchars);
+        if (data.compare(L"true") == 0)
+            return true;
+        if (data.compare(L"false") == 0)
+            return false;
+        throw std::runtime_error("Not a boolean.");
```
JeromeMartinez, line 131: I prefer silently quitting or return true.

JeromeMartinez (changes requested):
Thank you for you hard work on that.

In general this is OK but I would like some changes:

First I would like another PR with .gitignore, $(BDS) to ..\\.. stuff, MSVC2019 to MSVC2022 stuff, BCB22 to BCB23 stuff (if relevant? we don't use version 12), Manifest.rc/xml deletion (not used?), Exec regsvr32 to ExecWait regsvr32.exe, Delete "$INSTDIR\MediaInfo_InfoTip.dll" to  UnInstallLib, Refresh File Explorer.
It seems not related to the Win11 integration and I could independently validate them.

Then for this PR:
PNG should go in `\Source\Resource\Image\Assets`.

Please add the following header to all (non automatically generated and not tiny helper) .cpp & .h files:

``` 
/*  Copyright (c) MediaArea.net SARL. All Rights Reserved.
 *
 *  Use of this source code is governed by a BSD-style license that can
 *  be found in the License.html file in the root of the source tree.
 */
```

Source/GUI/VCL/GUI_Preferences.cpp
cjee21: This registry key is read by the shell extension to enable hiding it.

Source/GUI/VCL/GUI_Preferences.cpp
cjee21: No idea. You can try implement it later but we cannot perform operations that take too long there or else Explorer menu may hang. Also what if user highlights combination of files and folders?

Source/WindowsShellExtension/dllmain.cpp
cjee21: Going to have issues if both are installed at the same time. There will be two entries but both launch the same one.

Source/WindowsShellExtension/dllmain.cpp
cjee21: The exception is caught at the place which calls this function. It is silently handled there.

Source/GUI/VCL/GUI_Preferences.cpp
JeromeMartinez: I see that with Qt you check shellExtension and not hide-something, I don't see why it is not possible to do the same here and why another key is needed. 

i.e.:

```
#ifdef MEDIAINFO_QT
            if (!RegGetBool(HKEY_CURRENT_USER, L"Software\\MediaArea.net\\MediaInfo", L"shellExtension"))
#else
            if (RegGetDword(HKEY_CURRENT_USER, L"Software\\MediaArea\\MediaInfo", L"HideExplorerContextMenu"))
#endif // MEDIAINFO_QT
```

Replaced by:
```
#ifdef MEDIAINFO_QT
            if (!RegGetBool(HKEY_CURRENT_USER, L"Software\\MediaArea.net\\MediaInfo", L"shellExtension"))
#else
            if (!RegGetDword(HKEY_CURRENT_USER, L"Software\\MediaArea\\MediaInfo", L"ShellExtension"))
#endif // MEDIAINFO_QT
```

Source/Common/Preferences.cpp
cjee21: ShellExtension and ShellExtension_Folder is used multiple times here:
https://github.com/MediaArea/MediaInfo/blob/master/Source/Common/Preferences.cpp

So I did not want to touch that and break the old menu or mess up some registry.

Source/GUI/VCL/GUI_Preferences.cpp
JeromeMartinez: > You can try implement it later but we cannot perform operations that take too long there or else Explorer menu may hang.

I don't get there how the selection is done depending on the file extension. If we check the file extension we could check if it is a folder.
But low priority, I can accept such limitation for a first shot.

> Also what if user highlights combination of files and folders?

Same as if files are with a supported extension or not, IMO. Checking current Win10 behavior it is hidden if one of the selected items is not supported.

Source/Common/Preferences.cpp
JeromeMartinez: > So I did not want to touch that and break the old menu or mess up some registry.

My understanding is that we could live with (nearly) nothing changed in MI GUI so no risk to mess up.
I don't see why it is not possible to do https://github.com/MediaArea/MediaInfo/pull/960#discussion_r1844972395.

Source/GUI/VCL/GUI_Preferences.cpp
cjee21: Software\\MediaArea\\MediaInfo\\ShellExtension does not exist currently. Therefore the VCL code needs to write the registry anyway. It's just the naming. I named it hide because after install, there will be no registry entries. So if hide key does not exist then the context menu entry is not hidden. So that context menu is available immediately after install.

cjee21: I was wrong, I thought currently it is available after install that's why. I'll change it. Forget about making it available before first launch, not necessary when using the installer interactively as it auto launches after install.

cjee21: The Qt one will have it enabled by default right after install though.

Source/GUI/VCL/GUI_Preferences.cpp
JeromeMartinez: > At the moment it is shown for all file types.

I am reluctant to that. People will complain.
Is it difficult to add a list?
Based on [this list](https://github.com/MediaArea/MediaInfo/blob/master/Source/Install/MediaInfo_Extensions.nsh).

cjee21: Not sure if things like `audio` and `video` are allowed.

Source/GUI/VCL/GUI_Preferences.cpp
JeromeMartinez: > The Qt one will have it enabled by default right after install though.

I didn't see that but I am fine also with that, the registry key could be created also in the NSI.
But again, I prefer to split the PR in several smaller PRs, with the minimum in each PR.

Source/GUI/VCL/GUI_Preferences.cpp
JeromeMartinez: > Not sure if things like audio and video are allowed.

It would be even better, with the difficulty that some extensions are not recognized by Windows as audio or video.

Source/GUI/VCL/GUI_Preferences.cpp
cjee21: > the registry key could be created also in the NSI

That will only work for current user. But again, not necessary to have it by default in VCL version as mentioned above as it will auto launch after interactive install and auto open Pref window with shell extension checked by default. Qt one is different because it does not launch the Pref window on first launch and may not auto launch after install too.

Source/GUI/VCL/GUI_Preferences.cpp
cjee21: https://learn.microsoft.com/en-us/uwp/schemas/appxpackage/uapmanifestschema/element-desktop5-itemtype

Here are the docs. Looks like no `audio` and `video` allowed. So need to make sure every audio and video extension is manually specified.

Source/GUI/VCL/GUI_Preferences.cpp
JeromeMartinez: OK. I don't comment more the current PR, I let you change it / split it based on the discussion (I will quickly test and merge the PR which contains fixes not related to the Win11 integration so you don't have to maintain many branches at the same time).

Source/GUI/VCL/GUI_Preferences.cpp
JeromeMartinez: OK, so nee to fill all supported extensions.
And I see the special name `Directory` ;-).

cjee21:
> MSVC2019 to MSVC2022 stuff

A question, the project files for the new DLLs are only in MSVC2022. In that case how to go about in the installer script?

> Then for this PR: PNG should go in `\Source\Resource\Image\Assets`.

I put it there so that is is easier to re-generate the .pri file because everything has to be in the correct folder structure as what is in a MSIX. Do you still want to move it?

cjee21:
> Manifest.rc/xml deletion (not used?)

It is not currently used by C++Builder, a default manifest from C++Builder's directory is used. I deleted it in the same commit as adding `Manifest.manifest` so that there remains only the in-use manifest to prevent confusion.

> $(BDS) to ..\..

This is replacing the default manifest from C++Builder's directory with the new `Manifest.manifest` for package identity.

Source/GUI/VCL/GUI_Preferences.cpp
cjee21: Done with this. Generated with AI since it will be tedious to write that much. Not yet tested. Hope AI did not make any mistakes and that it works.

cjee21:
Done most of the changes discussed. I've totally removed changes with the GUI. I still have no clear idea how to go about it so you can give it a look/try.

JeromeMartinez:
>  I deleted it in the same commit as adding Manifest.manifest so that there remains only the in-use manifest to prevent confusion.

If I understand correctly this is independent from the Win11 context menu and useful as standalone. If so please a separate PR.

> A question, the project files for the new DLLs are only in MSVC2022. In that case how to go about in the installer script?

It seems that we have a coherency problem, we should go with MSVC2022 everywhere, we check. cc @g-maxime.

> I put it there so that is is easier to re-generate the .pri file because everything has to be in the correct folder structure as what is in a MSIX. Do you still want to move it?

I prefer to have all MediaInfo logs at the same place when not to complicated, and also I don't like to have the .pri (a binary file?) if not source in the source code. How is built the .pri? Would it make sense to have the logo files in the `Source\Resource\Image\Assets` and have a cmd (referenced in the project) which copy the file to the expected place and build the .pri?
Not a big deal if you prefer to keep it as is, we will check later if we can improve the build if you prefer to keep as is.

cjee21:
> > I deleted it in the same commit as adding Manifest.manifest so that there remains only the in-use manifest to prevent confusion.
> 
> If I understand correctly this is independent from the Win11 context menu and useful as standalone. If so please a separate PR.

The manifest is for granting package identity to MediaInfo GUI. It goes along with the sparse package.

> Would it make sense to have the logo files in the `Source\Resource\Image\Assets` and have a cmd (referenced in the project) which copy the file to the expected place and build the .pri? Not a big deal if you prefer to keep it as is, we will check later if we can improve the build if you prefer to keep as is.

The .pri only needs to be rebuilt when there is a change to the resources. If you prefer to rebuild it everytime, I will let you make the change. Guide: https://learn.microsoft.com/en-us/windows/msix/desktop/desktop-to-uwp-manual-conversion#generate-a-package-resource-index-pri-file-using-makepri

JeromeMartinez:
> The manifest is for granting package identity to MediaInfo GUI. It goes along with the sparse package.

OK.

> The .pri only needs to be rebuilt when there is a change to the resources. If you prefer to rebuild it everytime, I will let you make the change. 

OK, let's keep as is for the moment, You already do a lot and we'll manage the rebuild on our side (no need to be automatic, usually it is done by a check if the source files are younger than the binary, anyway we'll manage that).

***

I remark only now that we still build with MSVC2019 on our build farm, we'll migrate in order to be able to test this PR.

JeromeMartinez (commented):
Fine with that for the moment, but blocked because we don't have MSVC2022 on our build farm.

JeromeMartinez (approved):
Fine with that for the moment, but blocked because we don't have MSVC2022 on our build farm.

JeromeMartinez:
> The manifest is for granting package identity to MediaInfo GUI. It goes along with the sparse package.

I change my mind: my understanding now is that it does not hurt to switch to this manifest even without building the sparse package.
If so, please a separate PR with the BCB project modification + manifest, and I merge and build with this change, so this part is ready without waiting for the merge of this PR.

Source/GUI/VCL/GUI_Preferences.cpp
cjee21: @JeromeMartinez I was not wrong after all. The current MediaInfo context menu is enabled after install without requiring MediaInfo to be run at all. This is after an uninstall and install.

JeromeMartinez: My mistake. But let's keep it simple here anyway, for a first shot, or maybe we can find something simpler e.g. registering the Registry key in the installer and launching the DLL which checks this value?

cjee21:
Okay, tested the current state of this branch:
- The menu now only appears on the specified extensions + folders.*
- If multiple combinations are selected such as media + non-media + folder etc, it will appear if the file you right click on is a specified extension even if non-media are among the selected.*
- The old context menu is still shown so duplicate.^
- The GUI Preferences only controls the old context menu.^

\* = effect of updated manifest
^ = effect of removing the GUI changes

JeromeMartinez:
> The old context menu is still shown so duplicate.^
The GUI Preferences only controls the old context menu.^

Theses ones were not expected.
I see code in `WindowsShellExtension` handling the registry key. Isn't it possible to just launch `WindowsShellExtension` when the preference window is closed?
My concern is to avoid creating more registry keys and touch as few things as possible, not to remove everything.

cjee21:
> Theses ones were not expected. I see code in `WindowsShellExtension` handling the registry key. Isn't it possible to just launch `WindowsShellExtension` when the preference window is closed? My concern is to avoid creating more registry keys and touch as few things as possible, not to remove everything.

Let me try to explain in simple way how it works. 

1. The new DLL for shell extension needs to be registered as a context menu handler.
   - This can be done either by manually adding registry entries or by declaring it in the manifest of a MSIX.
   - MSIX method requires Windows build 17763 and above but since this is an unpackaged app, it requires sparse package which needs Windows build 19000 and above.
   - Registry manual method should work on Windows 7 or maybe even Vista onwards.
   - MSIX method is a must for the context menu entry to appear in the Windows 11 modern style context menu.
   - For now in this PR, only implement the MSIX method and only for Windows 11.

2. Once registered, when files are right-clicked in Explorer, Explorer will check list of registered handlers for that file extension. It will load the DLL in COM Surrogate / DLLHost (because it does not trust third party DLLs that may crash). This can be observed using Process Explorer from Sysinternals.

3. Then it calls the functions in our DLLMain.cpp after calling the entrypoint to get the COM interface.
    - GetState is called to determine if the menu entry should appear
      - In here, we check the registry that was set by GUI. If registry key indicates that it is disabled, we set cmdState = ECS_HIDDEN and Explorer will hide our entry. Else setting to ECS_ENABLED will enable the entry. Take note we only read registry here, no writing. Also code in this function should be fast so that Explorer menu does not hang or timeout.
    - GetIcon and GetTitle for providing the icon and title shown in the menu
    - For others like GetTooltip, we return E_NOTIMPL as we didn't implement

4. When we click on the menu entry, Invoke function is called by Explorer.
    - In here, we obtain from Explorer and parse the list of files selected. It is converted to command line and then we use that command line to start-up MediaInfo.exe and then exit this function.

5. About 10 seconds after the context menu is dismissed, our DLL will be unloaded by File Explorer from DLLHost. This can be observed using Process Explorer from Sysinternals.

So what else needs to be done in this PR?
For the Qt version, there is no other shell extensions to deal with and the Qt GUI already writes all its Preferences settings to the registry. So all is well and working properly. We can ignore Qt one now.
For the VCL one, this is where things have to be done. What needs to be done that I not yet do is...
- to write `Software\\MediaArea\\MediaInfo - ShellExtension DWORD = 0` when Preferences for shell is disabled by user and delete `ShellExtension DWORD` if enabled. Delete instead of setting to 1 since it is enabled by default so most people likely have it enabled so we avoid leaving registry keys behind since this is per user and can't be removed by uninstaller.
- to disable/remove the old context menu when new one is installed so that there are no duplicate entries in the 'classic' Explorer context menu,
- do something about the difference in behaviour of the two shell extensions
  - old one can toggle directories separately while new one can't
  - new one can open multiple files in same MediaInfo instance while old one launches one instance per file selected.

cjee21:
The problem here is I do not really understand how the existing context menu works and have no clear idea how to go about this. So in the meantime I remove all changes to GUI code so you can have a look and start with existing condition. I think you should have a better idea of how to proceed since you made the existing system.

cjee21:
Took a look at the codes for handling old context menu again and came up with a better solution. Appears to work properly on Windows 11 now. No duplicate entries and can control appearance of the new shell extension. Just need to test on Windows 10 and portable to ensure there is no change to existing behaviour there.

cjee21:
Note: we should consider adding notifying Windows Shell to refresh list of context menu upon installation/uninstallation. Else both the new and old menu entries will not appear after installation until Windows refreshes itself or something triggers a refresh althoough they are enabled by default.

Also, this bug: https://sourceforge.net/p/mediainfo/bugs/1161/ is reproducible and is now causing duplicate entries in 'classic' context menu for mp3 files etc.

cjee21:
Force push adds avif, avs, heic, heif, iamf, ico, jxl and webp to extensions list in manifest and changes webm to lower case.

Source/WindowsShellExtension/dllmain.cpp
```diff
@@ -0,0 +1,299 @@
...
+
+    IFACEMETHODIMP EnumSubCommands(_Outptr_ IEnumExplorerCommand** enumCommands) {
+        *enumCommands = nullptr;
+        return E_NOTIMPL;
+    }
+
+    IFACEMETHODIMP Invoke(_In_opt_ IShellItemArray* items, _In_opt_ IBindCtx* bindCtx) {
+        // Process items passed by File Explorer when context menu entry is invoked
+        UNREFERENCED_PARAMETER(bindCtx);
+        if (items) {
```
JeromeMartinez, line 209:
```suggestion
        if (!items) {
            return S_OK;
        }
```


And the other part 4 spaces smaller, instead, no need to go up to end of the function if no `items`.

Source/WindowsShellExtension/dllmain.cpp
```diff
@@ -0,0 +1,299 @@
...
+
+    IFACEMETHODIMP GetCanonicalName(_Out_ GUID* guidCommandName) {
+        *guidCommandName = GUID_NULL;
+        return S_OK;
+    }
+
+    IFACEMETHODIMP GetState(_In_opt_ IShellItemArray* items, _In_ BOOL okToBeSlow, _Out_ EXPCMDSTATE* cmdState) {
+        // Provide state of File Explorer context menu entry
+        // Hide it if registry setting indicates that it should be disabled, else it is enabled
+        UNREFERENCED_PARAMETER(items);
```
JeromeMartinez, line 175: My understanding he is that you have the items name so you can check if it is a directory or not, and `ECS_HIDDEN` if there are only directories shell extension for directories is not activated, it does not seem too complicated.

Does it make sense?

Source/Install/MediaInfo_GUI_Windows.nsi
```diff
@@ -210,6 +222,16 @@ Section Uninstall
     ExecWait '"$INSTDIR\ffmpeg_plugin_uninst.exe" /S _?=$INSTDIR'
     Delete "$INSTDIR\ffmpeg_plugin_uninst.exe"
 
+  ${If} ${AtLeastWin11}
+    System::Call '"$INSTDIR\MediaInfo_PackageHelper.dll"::_Uninstall@0() ? u'
```
JeromeMartinez, line 226: We could do that too (also install) when the user change the preferences, it seems easy and no interaction at all if the user does not want the shell extension.

Source/WindowsShellExtension/dllmain.cpp
cjee21: Probably. Need to make sure it does not take too long or else need check okToBeSlow and I am not sure what will happen if okToBeSlow is false. Maybe you can add it later on.

JeromeMartinez:
> Took a look at the codes for handling old context menu again and came up with a better solution.

Thank you, it seems less complicated than the previous proposal.

> Also, this bug: https://sourceforge.net/p/mediainfo/bugs/1161/ is reproducible and is now causing duplicate entries in 'classic' context menu for mp3 files etc.
UPDATE: Seems this is caused by MediaInfo still associated with audio and video in registry. Something in the old codes is not deleting properly. So basically, all these years, the existing context menu actually cannot be disabled completely. It is weird that it affects only some files like mp3 though because others like flac is also classified as audio in Windows.

Ha, I see! But so weird, indeed. I don't see an easy solution... Maybe just removing `video` & `audio` if it does not work well enough, and we manage our own list completely.

Source/WindowsShellExtension/dllmain.cpp
JeromeMartinez: > Need to make sure it does not take too long 

A possibility is to launch the check only if there are few items.
Also, we stop the check when there is at least 1 file.
My guess is that slowness will be very rare, people usually don't select 100 folder, it is more a couple of files + folder so a file will be caught early.

> if okToBeSlow is false

In that case we may just ignore the check.

cjee21:
> Ha, I see! But so weird, indeed. I don't see an easy solution... Maybe just removing `video` & `audio` if it does not work well enough, and we manage our own list completely.

Is there a mistake here?
https://github.com/MediaArea/MediaInfo/blob/36ff2a881993c83b52722b11a5c11807df70dbef/Source/Common/Preferences.cpp#L860-L863

Compare with:
https://github.com/MediaArea/MediaInfo/blob/36ff2a881993c83b52722b11a5c11807df70dbef/Source/Common/Preferences.cpp#L871
and
https://github.com/MediaArea/MediaInfo/blob/36ff2a881993c83b52722b11a5c11807df70dbef/Source/Common/Preferences.cpp#L873

Different/missing path? Could that be cause of not deleting audio and video? I am really not experienced with manually editing registries for shell extension so no idea what is going on. This Windows 11 context menu is the first time I am doing anything to do with shell extensions.

cjee21:
also there is difference of `Folder` `Directory.Audio` and `Directory.Video` in installer scipt and Preferences.cpp. Something needs to be cleaned up to prevent reg entries being left behind on disabled/uninstall.

cjee21:
> I don't see an easy solution... Maybe just removing `video` & `audio` if it does not work well enough, and we manage our own list completely.

I agree with this. But issue now is making sure all unwanted registry entries are correctly removed on uninstall or upgrade of MediaInfo.

Source/Install/MediaInfo_GUI_Windows.nsi
cjee21: try removing a package from within an app that belongs to the package, deadlock of app waiting for uninstall to complete while Windows Installer waits for app to exit
also removing the package will make MediaInfo no longer have package name in Task Manager
and we are not supposed to install and uninstall MSIX packages just for enabling/disabling a shell extension, the package represents the entire app

JeromeMartinez:
> But issue now is making sure all unwanted registry entries are correctly removed on uninstall or upgrade of MediaInfo.

Would you mind to add a PR about it? It seems that you see better than me the issue there.

Source/WindowsShellExtension/dllmain.cpp
cjee21: Microsoft's developer wrote it like that. Changed now.

cjee21:
> > But issue now is making sure all unwanted registry entries are correctly removed on uninstall or upgrade of MediaInfo.
> 
> Would you mind to add a PR about it? It seems that you see better than me the issue there.

Screenshot 2024-11-18 203519

Looks like the GUI is trying the wrong key that's why fail to disable for `audio` and `video`. There is no such thing as `HKCU\SystemFileAssociations`

cjee21:
Also I think the old shell will be re-enabled everytime MediaInfo is updated for those who disabled it until they open MediaInfo.

cjee21:
@JeromeMartinez I merged all my currently open PRs to a branch and made a test build. Everything seems to be in order now. Both context menu entry appears automatically after install, old one gets disabled on first launch of MediaInfo, no duplicate entries after that, can enable/disable shell extension and after uninstall, no registry entries that are written by MediaInfo are left behind.

cjee21:
Just a rebase, so that can test properly without the other bugs.

Source/WindowsShellExtension/dllmain.cpp
cjee21: Putting the link of the doc for this here for reference: https://learn.microsoft.com/en-us/windows/win32/api/shobjidl_core/nf-shobjidl_core-iexplorercommand-getstate

cjee21:
If we were to use the same DLL shell extension for older Windows versions, this is probably the way to do it: https://learn.microsoft.com/en-us/windows/win32/shell/reg-shell-exts#example-of-an-extension-handler-registration

JeromeMartinez:
> If we were to use the same DLL shell extension for older Windows versions, this is probably the way to do it:

Thank you.
I am not sure I want to spend the time on that but I would be happy to get an update if you wish to do so.
For the moment, the important part is Win11 integration (including the tooltip), it is less important for older versions.

cjee21:
> Thank you. I am not sure I want to spend the time on that but I would be happy to get an update if you wish to do so.

I do not want to spend time on Windows versions that will soon be end-of-support too and risk breaking things. The only benefit for older Windows versions is opening multiple files in the same instance and maybe some simplification of old codes.

> For the moment, the important part is Win11 integration (including the tooltip), it is less important for older versions.

Context menu is working fine so far and can now be independently controlled for folders as well after the last commit. ToolTip is already working when I tested after seeing issue 809 that day.

PS No idea how Windows decides the order to show menu items in the modern context menu but MediaInfo seems to always be the first entry on my PC.

cjee21:
btw it is possible to sign the NSIS uninstaller too to prevent yellow Windows warning when executing uninstall
https://nsis.sourceforge.io/Docs/Chapter5.html#uninstfinalize

There's also a note at https://nsis.sourceforge.io/Signing_an_Uninstaller that says:
> Note 2: Running a dummy command with !uninstfinalize may also be useful if only the installer will be signed. This could prevent bogus header fields in the uninstaller, see [feature request ticket #577](https://sourceforge.net/p/nsis/feature-requests/577/)

JeromeMartinez:
> btw it is possible to sign the NSIS uninstaller too to prevent yellow Windows warning when executing uninstall

It was the case several years ago and at some point it disappeared, I was too lazy for finding the reason :(.

cjee21:
> > btw it is possible to sign the NSIS uninstaller too to prevent yellow Windows warning when executing uninstall
> 
> It was the case several years ago and at some point it disappeared, I was too lazy for finding the reason :(.

I mean currently, the uninstaller is not signed so when user execute uninstall, this appears instead of what we see during install:
Untitled

If we sign it and add MediaInfo icon then the same UAC window should appear like the one during install with publisher name and icon.

JeromeMartinez:
> I mean currently, the uninstaller is not signed so when user execute uninstall

I was speaking about that too :).

cjee21:
@JeromeMartinez managed to get a signed uninstaller.

Just need to add this to the `.nsi` files:
```
; Sign uninstaller
!uninstfinalize 'sign.cmd "%1"'
```

Then create a bat/cmd file named `sign` with the following:
```
set /P CodeSigningCertificatePass= < %USERPROFILE%\CodeSigningCertificate.pass
if "%NOSIGN%"=="" (
    signtool sign /f %USERPROFILE%\CodeSigningCertificate.p12 /p %CodeSigningCertificatePass% /fd sha256 /v /tr http://ts.ssl.com /td sha256 /d "MediaInfo Uninstaller" /du http://mediaarea.net %1
)
set CodeSigningCertificatePass=
```

After that just build as usual and the uninstaller should be signed. When user uninstalls, UAC will show MediaInfo Uninstaller with verified publisher instead of uninst.exe with unknown publisher.

cjee21:
Rebased to master.

NSIS needs to be updated to 3.10 (at least 3.09) to support the `${AtLeastWin11}` check used in this PR. Should consider updating the 8-year-old 7-Zip while at it. Maybe update libcurl too if there are security fixes.
https://github.com/MediaArea/MediaArea-Utils-Binaries/tree/master/Windows

cjee21:
@g-maxime 
I tested https://mediaarea.net/download/snapshots/binary/mediainfo-gui/20241128-2/MediaInfo_GUI_24.11.20241128_Windows_x64.exe

MediaInfo fails to start because `Source/GUI/VCL/Manifest.manifest` does not support capital V for `Version`. Other than that the new context menu installs and uninstalls fine. Also, for some reason the version numbers for the two new DLLs does not seem to be updated successfully.

cjee21:
> Also, for some reason the version numbers for the two new DLLs does not seem to be updated successfully.

Likely because the resource files were generated by Visual Studio and uses UTF-16 LE encoding.

JeromeMartinez:
@cjee21 thank you for the debugging.

JeromeMartinez:
> Likely because the resource files were generated by Visual Studio and uses UTF-16 LE encoding.

Does it hurt if we switch to UTF-8?

cjee21:
Okay let's try this. Converted to UTF-8 and simplified.

JeromeMartinez:
> Okay let's try this. Converted to UTF-8 and simplified.

Thank you. Small thing but could you merge the commits adding the new files in UTF-16 and the UTF-8 switch, so we have only one commit for the addition of the new file and no UTF-16 intermediate.

Also, please remove the `<SpectreMitigation>Spectre</SpectreMitigation>` lines, for the moment we don't have it in other place, I think we would add it everywhere in one shot when we decide to switch to it.

g-maxime:
I've already relaunched a build, so the DLL version will not be updated in the next snapshot either.

JeromeMartinez:
> Time for an interactive rebase......

Thank you.
We are not far from the end for this PR, no other change request expected.

cjee21:
> Also, please remove the `<SpectreMitigation>Spectre</SpectreMitigation>` lines, for the moment we don't have it in other place, I think we would add it everywhere in one shot when we decide to switch to it.

There is also EH Continuation Metadata (EHCONT) that is not yet enabled elsewhere in MediaInfo.

Instruction Pointer Validation section of https://techcommunity.microsoft.com/blog/windowsosplatform/developer-guidance-for-hardware-enforced-stack-protection/2163340

JeromeMartinez:
> There is also EH Continuation Metadata (EHCONT) that is not yet enabled elsewhere in MediaInfo.

I learn new option every day.
But I prefer to have a common set of options in all project, and debate about options in separate PRs.

cjee21:
https://mediaarea.net/download/snapshots/binary/mediainfo-gui/20241128-3/MediaInfo_GUI_24.11.20241128_Windows_x64.exe

Appears to be working fine 😃 

Screenshot 2024-11-28 225009
Screenshot 2024-11-28 225133

JeromeMartinez:
@cjee21 forget about my previous comment, it is working as expected, there is just a delay between the install and the menu displayed.

cjee21:
> there is just a delay between the install and the menu displayed.

There is a shell refresh triggered after provisioning of package by installer but at that point the package is not installed by Windows yet. Since it is uncertain when Windows will finish installing the provisioned package, I decide to just let File Explorer refresh itself which seems to happen within about 30 seconds of any change in installed packages. There will be duplicate entries in the "show more options" after install as the legacy entry will be removed on first launch of MediaInfo when it checks that the new one is successfully registered.

cjee21:
https://mediaarea.net/download/snapshots/binary/mediainfo-gui/20241129-2/

The 32/64 combined installer has a bug. Some files are wrongly placed in assets folder. Think I overlooked something in the script regarding `SetOutPath`.

Other things look good. Version numbers are correct and updating from previous snapshot correctly updates the package.

PR #965, opened 2024-11-16
Qt GUI: Update MSIX manifest

cjee21:
Lower Windows version requirement and add list of file extensions for FileExplorerContextMenus.

### DIFF
--- a/Source/WindowsQtPackage/AppxManifest.xml
+++ b/Source/WindowsQtPackage/AppxManifest.xml
@@ -18,7 +18,7 @@
     <Resource Language="en-us" />
   </Resources>
   <Dependencies>
-    <TargetDeviceFamily Name="Windows.Desktop" MinVersion="10.0.19000.0" MaxVersionTested="10.0.26100.0" />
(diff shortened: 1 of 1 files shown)


cjee21:
FYI with the current Qt MSIX build script, MSIX package is 149MiB and install size is 339MiB.

JeromeMartinez:
Nice to have that in the source code but:

>  MSIX package is 149MiB

It is definitely a blocker for general release, lot of people love to have MediaInfo small.
Maybe later ever removing the HTML engine or using the integrated one... But not a priority for the moment.

cjee21:
>Maybe later ever removing the HTML engine or using the integrated one... But not a priority for the moment.

Also can make VCLibs a dependency rather than statically linking or bundling the DLLs to reduce size further. Most systems should have VCLibs installed already and it should auto-install for those that don't if added as dependency. Will also ensure VCLibs is kept updated without needing to update the app.

If others are interested in Qt version, maybe can consider publishing to Store as an alternate / beta version.

PR #967, opened 2024-11-18
Update file extensions lists

cjee21:


JeromeMartinez (changes requested):
Adding a file extension should be done on all GUIs and platforms, please update at the same time `Source\Common\Preferences.cpp`, `Source\GUI\Cocoa\MediaInfo-Info.plist`, `Source\GUI\UWP\Package.appxmanifest`, `Source\Install\MediaInfo_Extensions.nsh`.

JeromeMartinez:
I saw https://github.com/MediaArea/MediaInfo/pull/960#issuecomment-2481973220 too late. Lists should be same everywhere.

cjee21:
> I saw [#960 (comment)](https://github.com/MediaArea/MediaInfo/pull/960#issuecomment-2481973220) too late. Lists should be same everywhere.

I will try to see if I can come up with a way to auto generate all the lists from a single list. To ensure all are the same and make it easier to edit the lists next time.

cjee21:
@JeromeMartinez Need to do something about the audio video thing.

In installer there is:
  !insertmacro MediaInfo_Extensions_Uninstall_I "audio"
  !insertmacro MediaInfo_Extensions_Uninstall_I "Folder"
  !insertmacro MediaInfo_Extensions_Uninstall_I "video"

In Preferences.cpp there is:
        ExplorerShell_Edit("Software\\Classes\\Directory", Config.Read(__T("ShellExtension_Folder")).To_int32s(), IsChanged);

So Folder and Directory is mismatch as well as Preferences.cpp list no audio and video. Also, the Preferences.cpp code is not correctly removing the audio and video entries added by installer when user chooses to disable the shell extension.

cjee21:
Done matching all except `Source\GUI\Cocoa\MediaInfo-Info.plist` since the list is split in there so I do not know what goes where and what is missing.

Here is the spreadsheet that can auto-generate the lists:
[MediaInfo_fileextensions.xlsx](https://github.com/user-attachments/files/17799301/MediaInfo_fileextensions.xlsx)

JeromeMartinez (approved):

JeromeMartinez:
> So Folder and Directory is mismatch as well as Preferences.cpp list no audio and video. Also, the Preferences.cpp code is not correctly removing the audio and video entries added by installer when user chooses to disable the shell extension.

Oops... If I understand correctly [this post](https://www.tenforums.com/general-support/205588-diff-hkey_current_user-software-classes-directory-vs-folder.html#post2548121), I should uninstall "Folder" and install "Directory", and let the CPP as is.

JeromeMartinez:
> So Folder and Directory is mismatch as well as Preferences.cpp list no audio and video. Also, the Preferences.cpp code is not correctly removing the audio and video entries added by installer when user chooses to disable the shell extension.

I should read all notifications before answering, now seeing https://github.com/MediaArea/MediaInfo/pull/968, thanks!

PR #968, opened 2024-11-18
Windows GUI: Fix shell extension registries

cjee21:
Fix bugs to prevent issues such as the following:
https://sourceforge.net/p/mediainfo/support-requests/51/
https://sourceforge.net/p/mediainfo/bugs/1161/ 

Make sure that associations other than for file extensions in MediaInfo's list are properly removed and no longer added.
Also refresh File Explorer's context menu after install/uninstall so changes are visible immediately.

Tests results:
- Context menu entry appears immediately and for both files and folders after install without requiring MediaInfo to be launched first regardless of MediaInfo's existing settings.
- When MediaInfo is launched, context menu is updated according to settings set.
- If shell extension is disabled, context menu entry no longer shows for files associated with audio/video such as .mp3.
- After uninstall, the registry was searched and it was found that some MediaInfo entries still remain. This should be retested after #967 is merged.

### DIFF
--- a/Source/Common/Preferences.cpp
+++ b/Source/Common/Preferences.cpp
@@ -867,10 +867,10 @@ int Preferences::ExplorerShell()
                 }
             }
         }
-        ExplorerShell_Edit("SystemFileAssociations\\audio", 0, IsChanged);
-        ExplorerShell_Edit("SystemFileAssociations\\Directory.Audio", 0, IsChanged);
-        ExplorerShell_Edit("SystemFileAssociations\\Directory.Video", 0, IsChanged);
-        ExplorerShell_Edit("SystemFileAssociations\\video", 0, IsChanged);
+        ExplorerShell_Edit("Software\\Classes\\SystemFileAssociations\\audio", 0, IsChanged);
+        ExplorerShell_Edit("Software\\Classes\\SystemFileAssociations\\Directory.Audio", 0, IsChanged);
+        ExplorerShell_Edit("Software\\Classes\\SystemFileAssociations\\Directory.Video", 0, IsChanged);
+        ExplorerShell_Edit("Software\\Classes\\SystemFileAssociations\\video", 0, IsChanged);
 
         //Adding/removing to SystemFileAssociations
         int32s ShellExtension=Config.Read(__T("ShellExtension")).To_int32s();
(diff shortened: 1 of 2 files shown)


JeromeMartinez:
> Context menu entry appears immediately and for both files and folders after install without requiring MediaInfo to be launched first regardless of MediaInfo's existing settings.

I didn't think to when user disabled it then install an update :(.
A nice to have in the NSI would be a test on the registry key, if it exist and 0, no extension installation/update (so install only at the first install)

> If shell extension is disabled, context menu entry no longer shows for files associated with audio/video such as .mp3.

Thank you.

> After uninstall, the registry was searched and it was found that some MediaInfo entries still remain. This should be retested after https://github.com/MediaArea/MediaInfo/pull/967 is merged.

Please rebase.

Source/Common/Preferences.cpp
```diff
@@ -857,10 +857,10 @@ int Preferences::ExplorerShell()
                 }
             }
         }
-        ExplorerShell_Edit("SystemFileAssociations\\audio", 0, IsChanged);
-        ExplorerShell_Edit("SystemFileAssociations\\Directory.Audio", 0, IsChanged);
-        ExplorerShell_Edit("SystemFileAssociations\\Directory.Video", 0, IsChanged);
-        ExplorerShell_Edit("SystemFileAssociations\\video", 0, IsChanged);
```
JeromeMartinez, line 863: Shouldn't we delete the MediaInfo related entries in case it is from a previous version?

JeromeMartinez (changes requested):

Source/Common/Preferences.cpp
cjee21: The previous code (now in red) is trying to remove entries from `HKCU\SystemFileAssociations` if I understand correctly? See https://github.com/MediaArea/MediaInfo/pull/960#issuecomment-2482925278. Since that key does not exist, nothing happens. That is why it fails to completely disable the shell ext previously. Now it is changed to delete the correct key but actually may not be necessary as the key is now deleted by NSIS on upgrade/install. I am not sure if the key exists on portable. The one I have was created by NSIS.

JeromeMartinez: My mistake. 

>  I am not sure if the key exists on portable. The one I have was created by NSIS.

Portable is the cpp file you are checking without the NSIS, so no diff.

PR #976, opened 2024-11-26
Windows GUI: Implement signing of uninstaller

cjee21:
Should make a test build before merging to ensure it works as expected on the build farm.


cjee21:
This one looks good. The uninstaller is signed and shows as verified by UAC on uninstall.
https://mediaarea.net/download/snapshots/binary/mediainfo-gui/20241126-6/MediaInfo_GUI_24.11.20241126_Windows_x64.exe

PR #978, opened 2024-11-27
Release scripts improvements

cjee21:
- Change quotes for `set BPATH` in Windows GUI/CLI release scripts to match what is done in `Release_GUI_Windows.bat`. This prevents failure if the path contains spaces or other characters that can cause issues.
- Remove `Preparing` section from `Release_GUI_Windows_i386.bat` as it is not present in `Release_GUI_Windows_x64.bat` and is not needed. The file that is attempted to copy/rename does not exist including its directory.

### DIFF
--- a/Release/Release_CLI_Windows_i386.bat
+++ b/Release/Release_CLI_Windows_i386.bat
@@ -8,8 +8,8 @@
 
 rem --- Search binaries ---
 set BPATH=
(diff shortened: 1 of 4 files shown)


JeromeMartinez:
Thanks.

JeromeMartinez:
[Snapshot](https://mediaarea.net/download/snapshots/binary/mediainfo-gui/20241127/).

PR #982, opened 2024-12-02
Qt GUI: Add Windows ARM64 build

cjee21:
Needs all the other zlib, ZenLib, MediaInfoLib and Qt PRs for ARM64 and EHCONT.

Tested to build and produce an ARM64 binary. However, Qt does not have pre-built ARM64 WebEngine and I do not have an ARM64 PC to test so this is not fully tested.

### DIFF
--- a/Project/QMake/GUI/MediaInfoQt.pro
+++ b/Project/QMake/GUI/MediaInfoQt.pro
@@ -66,6 +66,9 @@ unix {
 }
 
 win32 {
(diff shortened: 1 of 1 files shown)


cjee21:
Builds without errors after the other PRs were merged with WebEngine parts disabled as there are no pre-built ARM64 WebEngine from Qt and I did not manually build it.

Screenshot 2024-12-04 014125

PR #985, opened 2024-12-04
Windows GUI: Build for ARM64 PC

cjee21:
Depends on:
- [ ] https://github.com/MediaArea/MediaInfoLib/pull/2161

Note:
- If there is any condition where InfoTip needs libcurl then we need an ARM64X version of libcurl. If not, it should work as it is or we can use an ARM64EC version of libcurl for native code.
- The ARM64 is not added to all-in-one installer. It should be either added or there should be a way to direct ARM64 users to the correct installer or else they will have broken shell integration. Maybe time for an online stub installer?


cjee21:
> why not the at x64 installer?

Windows 10 does not have x64 emulation on ARM64. It is Windows 11 exclusive.

cjee21:
> a check that it is not Windows 10 (and leave if it is) should be added to the ARM64 installer, shouldn't it?

ARM64 installer checks for ARM64 and Windows 11 oninit and aborts if any condition not met. Should be good enough?

https://github.com/cjee21/MediaInfo/blob/033545d35b196d92e8ac9686b76f9850dd7ffd1d/Source/Install/MediaInfo_GUI_Windows_ARM64.nsi#L123-L132

JeromeMartinez:
> ARM64 installer checks for ARM64 and Windows 11 oninit and aborts if any condition not met. Should be good enough?

It is! I missed it, thanks for having done it before I ask.

cjee21:
> I edited my proposal. Windows 10 ARM64 shell extension support is not a big need IMO, @cjee21 only if you are super motivated to do it...

Very low market share. Probably extinct soon. So nah... :p

Or we can get the Qt x64 and ARM64 to build and post it to the Store replacing the current very outdated UWP version (I got 2022 version from store?) and it will serve Windows 10 ARM64 as well as users who can only install apps from Store. Qt one is now quite close to VCL in features compared to UWP/wx. It is also probably easier to put Qt one in Store as it does not attempt to write registry entries for shell extensions.

JeromeMartinez:
> Very low market share. Probably extinct soon. So nah... :p

OK, so let's keep with the i386 offer in the case of Windows 10 ARM64 and we don't really care.

> Or we can get the Qt x64

Another topic but for the moment I am more in the mood of keeping VCL. It may change, I still need to think about it. But most of the remaining work does not depend on that, so you can go with it whatever is my future choice.

JeromeMartinez:
> Very low market share. Probably extinct soon. So nah... :p

I edited again my proposal and now it is relatively small amount of remaining work, @cjee21 would you mind to manage that?

cjee21:
The Qt x64 one is more or less ready for the Store. The fully native Qt ARM64 one needs an ARM64 build of Qt WebEngine and Graph/FFmpeg plugins which is not something I can do. That's all for Qt.

> I edited again my proposal and now it is relatively small amount of remaining work, @cjee21 would you mind to manage that?

Remains the stub installer right? I am not sure if I can manage to do that. Firstly, it requires knowing the URLs of the installers. Second, I just started doing NSIS this year and have no idea how to make a downloader yet. 😅

JeromeMartinez:
> Firstly, it requires knowing the URLs of the installers. Second, I just started doing NSIS this year and have no idea how to make a downloader yet. 😅

Not difficult but fair, you already did a lot, we'll check.

cjee21:
> But at some point we may want to let the user enter a URL in the GUI (feature requested by several users) so we'll need to add libcurl to the DLL called by the GUI, we'll do it.

@JeromeMartinez I pasted a url in the "File name" field of Windows File open dialog and it loads in MediaInfo 😆 

Screenshot 2024-12-04 210902

Screenshot 2024-12-04 210920

cjee21:
> It is hidden for the moment, people usually don't try that and ask for a more obvious interface.

Tried on Qt one and it appears to work but since Qt one produces blank output when passing urls in command line, it also produces blank using the file open dialog.

JeromeMartinez:
> I tried making a full stub installer and it appears to work.

Go, PR :).
The hybrid stub version should be even easier.

cjee21:
Checked https://mediaarea.net/download/snapshots/binary/mediainfo-gui/20241205-2/MediaInfo_GUI_24.11.20241205_Windows_ARM64.exe

Looks okay other than `MediaInfo_WindowsShellExtension.dll` not digitally signed.

cjee21:
@JeromeMartinez So left the testing of ARM64 installer and after that if it is working fine, dealing with the combined installer and making a version-less stub.

JeromeMartinez:
@rednoah please test [this ARM build](https://mediaarea.net/download/snapshots/binary/mediainfo-gui/20241205-3/MediaInfo_GUI_24.11.20241205_Windows_ARM64.exe), especially the Windows Explorer integration (shell extension and tooltip).

cjee21:
> @rednoah please test [this ARM build](https://mediaarea.net/download/snapshots/binary/mediainfo-gui/20241205-3/MediaInfo_GUI_24.11.20241205_Windows_ARM64.exe), especially the Windows Explorer integration (shell extension and tooltip).

Link no longer exists. Try this one: [MediaInfo_GUI_24.11.20241205_Windows_ARM64.exe](https://mediaarea.net/download/snapshots/binary/mediainfo-gui/20241205/MediaInfo_GUI_24.11.20241205_Windows_ARM64.exe)

rednoah:
Nice work! I can confirm that it works on my `arm64` machine.

menu
tooltip

PR #986, opened 2024-12-04
Windows GUI: Add online/stub installer

cjee21:
There is still room for improvement/refinement...

### DIFF
--- a/Release/Release_GUI_Windows.bat
+++ b/Release/Release_GUI_Windows.bat
@@ -23,6 +23,7 @@ del   MediaInfo_GUI_Windows_i386.exe
 rem --- Installer ---
 pushd %BPATH%\Windows\NSIS
 makensis.exe "%~dp0\..\Source\Install\MediaInfo_GUI_Windows.nsi"
+makensis.exe "%~dp0\..\Source\Install\MediaInfo_GUI_Windows_Stub.nsi"
 makensis.exe "%~dp0\..\Source\Install\MediaInfo_GUI_Windows_x64.nsi"
 makensis.exe "%~dp0\..\Source\Install\MediaInfo_GUI_Windows_ARM64.nsi"
 makensis.exe "%~dp0\..\Source\Install\MediaInfo_GUI_Windows_i386.nsi"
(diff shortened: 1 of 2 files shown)


JeromeMartinez:
Thank you.
I am thinking to have a stub installer without the version so not rebuilt at each version, and the version would be available online.

cjee21:
> Thank you. I am thinking to have a stub installer without the version so not rebuilt at each version, and the version would be available online.

That is what I am thinking too (like other apps [Firefox?] so users can reuse previously downloaded installer too) but that requires a special URL so I cannot do that now right?

JeromeMartinez:
> but that requires a special URL so I cannot do that now right?

Right. We need to discuss internally about that. I'll let you know.
Also, a stub by version is also good, version is used by some third party websites for specific the possibility to use a specific version, so it would be an addition.

cjee21:
Need find a way to pass language to the 2nd installer if possible. Right now on first time install (not updating) user needs to select language two times and the 2nd installer appears behind the 1st after language selection.

cjee21:
Simplified... less than 90KiB now and no double installer visible :)

JeromeMartinez:
@g-maxime more stuff for the version upgrade script.

cjee21:
> @g-maxime more stuff for the version upgrade script.

and digitally signing another installer exe.

cjee21:
@JeromeMartinez If we want to make the stub installer more secure, we can add a check that the downloaded installer is digitally signed with the correct cert before executing it. It will require third party plugin such as [the one from Mozilla](https://searchfox.org/mozilla-central/source/other-licenses/nsis/Contrib/CertCheck) used by Firefox's stub installer or [another one](https://github.com/past-due/nsisfilecheck).

If the version-less stub is still desired, one possibility is to add a config or ini file containing version and/or URL to the server that can be read by NSIS [ConfigRead](https://nsis.sourceforge.io/ConfigRead) or [ReadINIStr](https://nsis.sourceforge.io/Reference/ReadINIStr) functions.

JeromeMartinez:
> If the version-less stub is still desired, one possibility is to add a config or ini file containing version and/or URL to the server that can be read by NSIS [ConfigRead](https://nsis.sourceforge.io/ConfigRead) or [ReadINIStr](https://nsis.sourceforge.io/Reference/ReadINIStr) functions.

We have already a lot of places with the version, we'll check in the team where the version-less installer should check that.

JeromeMartinez:
> and digitally signing another installer exe.

All is fine now, right?

PR #987, opened 2024-12-05
Windows GUI: Ensure correct installers are used for each system

cjee21:
Since there is now ARM64.


JeromeMartinez:
For coherency, not something similar for i386 (test on i386 only or ARM64 on Win10)?

cjee21:
For combined installer will be in separate PR along with adding ARM64 handling after making sure ARM64 works fine.

JeromeMartinez:
> after making sure ARM64 works fine.

Exactly.

cjee21:
@JeromeMartinez adding ARM64 to combined installer increases size by roughly 3.5MiB (excluding digital signatures). I think it is acceptable. If want to really save bandwidth can use the online installer and only use combined when offline install is needed.

PR #990, opened 2024-12-06
Handle snapshots in stub installer

g-maxime:


Source/Install/MediaInfo_GUI_Windows_Stub.nsi
```diff
@@ -49,14 +54,14 @@ Section
   ${If} ${IsNativeARM64}
     ${AndIf} ${AtLeastWin11}
       inetc::get /CAPTION "MediaInfo Online Installer" /BANNER "Downloading MediaInfo ${PRODUCT_VERSION} ARM64..." \
-      "https://mediaarea.net/download/binary/mediainfo-gui/${PRODUCT_VERSION}/MediaInfo_GUI_${PRODUCT_VERSION}_Windows_ARM64.exe" "$PLUGINSDIR\MediaInfoInstaller.exe"
+      "${BASEURL}/${PRODUCT_VERSION}/MediaInfo_GUI_${PRODUCT_VERSION}_Windows_ARM64.exe" "$PLUGINSDIR\MediaInfoInstaller.exe"
```
cjee21, line 57: `${PRODUCT_VERSION}/` should not be here like for the other archs below.

cjee21: Also, since `BASEURL` already ends with a slash, will the double slash cause issues?

PR #991, opened 2024-12-09
Windows GUI: Add evergreen online/stub installer

cjee21:
Add a version-less evergreen online/stub installer that will install the latest available suitable version for all Windows versions and architectures.

Also update the versioned online/stub installer to be consistent with the version-less one and simplify it.

### DIFF
--- a/Source/Install/MediaInfo_GUI_Windows_EvergreenStub.nsi
+++ b/Source/Install/MediaInfo_GUI_Windows_EvergreenStub.nsi
@@ -1,0 +1,147 @@
+#NSIS: encoding=UTF-8
+RequestExecutionLevel admin
+ManifestDPIAware true
+
+; Some defines
+!define PRODUCT_NAME "MediaInfo"
+!define PRODUCT_PUBLISHER "MediaArea.net"
+!define PRODUCT_VERSION "24.12"
+!define PRODUCT_VERSION4 "${PRODUCT_VERSION}.0.0"
+
+; Compression
+SetCompressor /FINAL /SOLID lzma
+
+; x64 stuff
+!include x64.nsh
+
+; Logic stuff
+!include LogicLib.nsh
+
+; Windows version stuff
+!include WinVer.nsh
+
+; String functions
+!include StrFunc.nsh
+${Using:StrFunc} StrTrimNewLines
+
+; Word functions
+!include WordFunc.nsh
+
+; Installer icon
+Icon "..\..\Source\Resource\Image\MediaInfo.ico"
+
+; Enable Visual Styles
+XPStyle on
+
+; Info
+VIProductVersion "${PRODUCT_VERSION4}"
+VIAddVersionKey /LANG=0 "CompanyName"      "${PRODUCT_PUBLISHER}"
+VIAddVersionKey /LANG=0 "ProductName"      "${PRODUCT_NAME}"
+VIAddVersionKey /LANG=0 "ProductVersion"   "${PRODUCT_VERSION4}"
+VIAddVersionKey /LANG=0 "FileDescription"  "All about your audio and video files"
+VIAddVersionKey /LANG=0 "FileVersion"      "${PRODUCT_VERSION4}"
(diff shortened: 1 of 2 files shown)


Source/Install/MediaInfo_GUI_Windows_EvergreenStub.nsi
```diff
@@ -0,0 +1,106 @@
+#NSIS: encoding=UTF-8
+RequestExecutionLevel admin
+
+; Some defines
+!define PRODUCT_NAME "MediaInfo"
+!define PRODUCT_PUBLISHER "MediaArea.net"
+!define PRODUCT_VERSION "2024.12.9"
```
JeromeMartinez, line 7: I understand the proposal here but I think I prefer to keep consistency and use 24.12 here.

JeromeMartinez (changes requested):

Source/Install/MediaInfo_GUI_Windows_EvergreenStub.nsi
cjee21: Done.

JeromeMartinez (approved):

cjee21:
Note: the [evergreen online/stub installer](https://github.com/MediaArea/MediaInfo/pull/991/commits/9af4adf24acc0c6660395602846024a3bb30c498) is not working as intended because https://mediaarea.net/mediainfo_check/version.txt is returning [changelog.bin](https://github.com/MediaArea/MediaArea-Website/blob/master/web/mediainfo_check/changelog.bin) instead of [version.txt](https://github.com/MediaArea/MediaArea-Website/blob/master/web/mediainfo_check/version.txt).

cjee21:
Update: the [evergreen online/stub installer](https://github.com/MediaArea/MediaInfo/pull/991/commits/9af4adf24acc0c6660395602846024a3bb30c498) is now working as intended since the server is now serving the expected file.

So the only thing is whether we want to add signature verification. If yes, need to add the third party NSIS plugin DLL to the repository containing NSIS first.

JeromeMartinez:
> So the only thing is whether we want to add signature verification.

I like the maximum balance between security, file size and dev time, if you consider that it is fine for you to add it I take it ☺️.

> If yes, need to add the third party NSIS plugin DLL to the repository containing NSIS first.

Which NSIS plugin DLL?

cjee21:
> Which NSIS plugin DLL?

As I mentioned earlier [here](https://github.com/MediaArea/MediaInfo/pull/986#issuecomment-2527314143), there are two possible plugins that I know of that should work. As far as I know, NSIS does not have built-in signature verification features so third party plugin is needed. The plugins basically calls Windows built-in verification features.

JeromeMartinez:
> The plugins basically calls Windows built-in verification features.

It helps about maintenance.
I have a preference for the Mozilla one, easy to add to the source code and we built it.
Can you start without our repo having the binary?

cjee21:
> > The plugins basically calls Windows built-in verification features.
> 
> It helps about maintenance. I have a preference for the Mozilla one, easy to add to the source code and we built it. Can you start without our repo having the binary?

I can try to compile that and add the verification to our script and see if it works first.

cjee21:
Done adding digital signature verification using the Mozilla CertCheck plugin.

Both online installers are still under 100KiB.

The source code for the plugin can be obtained from:
https://github.com/mozilla/gecko-dev/tree/master/other-licenses/nsis/Contrib/CertCheck

The pre-built and digitally signed plugin DLL can be obtained from:
https://github.com/mozilla/gecko-dev/blob/master/other-licenses/nsis/Plugins/CertCheck.dll

cjee21:
@JeromeMartinez Since this was merged before `CertCheck.dll` is added to https://github.com/MediaArea/MediaArea-Utils-Binaries/tree/master/Windows/NSIS/Plugins/x86-unicode, I suppose NSIS build for online installer will fail and the build script may then fail due to how it is written?

Issue #994, opened 2024-12-13
Windows 11 integration need an option for keeping the old context menu


JeromeMartinez:
Feedback from a user.
When calling the context menu from Total Commander, there is no more the context menu because the Windows 11 Explorer style is not supported, so we need an option for keeping the old context menu as it is done in other tools.

WinRAR:
![2024-12-12_171853](https://github.com/user-attachments/assets/094f666d-1990-4a36-9cb6-ba782791994c)

TreeSize:
![2024-12-12_173110](https://github.com/user-attachments/assets/c1b80c48-de6e-42f4-8b10-2d1be99743b8)

cjee21:
Unable to reproduce. Windows 11 shell extensions work fine in Total Commander including properly opening the file in MediaInfo and Mp3tag.
The entries for Clipchamp, MediaInfo, Quick Share, Mp3tag and Notepad are all Windows 11 new-type shell extensions.

![Screenshot 2024-12-13 201113](https://github.com/user-attachments/assets/fda180bb-0978-4175-8664-0d8d1343d9ab)

![Screenshot 2024-12-13 201539](https://github.com/user-attachments/assets/105c77a4-2336-4a20-ae12-baa65e58f42d)

In my opinion the user should contact developer of Total Commander for any compatibility issues instead of requesting changes from each and every app to be compatible with Total Commander. Most apps that implement the new shell extension no longer have the legacy shell extension. 

Unrelated note: Many apps do not even have an option to disable the context menu entry. Mp3tag only has option to disable in Store version while the installer version can only enable/disable the context menu during installation.

sevastopolcat:
Why is this so? WinRAR, EZCDAC & Mp3tag - from main Win11 menu.

![TC](https://github.com/user-attachments/assets/bbf525f4-338f-4eec-bc97-a99092996418)

cjee21:
Are both options enabled in MediaInfo Preferences?
![Screenshot 2024-12-13 233135](https://github.com/user-attachments/assets/bfcf2635-5db8-4c98-9463-e71ee05f290a)

sevastopolcat:
Yes

![2024-12-13_183356](https://github.com/user-attachments/assets/4fdd741d-6a12-4dd1-a823-d912ba671bea)

cjee21:
If it works properly in Windows File Explorer but not on Total Commander then I have no idea why. It works when I tested in Total Commander. Only the developer of Total Commander will know.

[...]

PR #997, opened 2024-12-16
Rename shell extension Verb Id

cjee21:
Workaround Windows issue causing shell extension to not appear in classic context menus on certain systems.
As seen in https://github.com/MediaArea/MediaInfo/issues/994#issuecomment-2541700900 and https://github.com/MediaArea/MediaInfo/issues/994#issuecomment-2541738716.

I have no idea what letter works in our case so let us just try "C" since it worked for NanaZip.

Thanks to https://github.com/M2Team/NanaZip/pull/205.

### DIFF
--- a/Source/WindowsSparsePackage/MSIX/AppxManifest.xml
+++ b/Source/WindowsSparsePackage/MSIX/AppxManifest.xml
@@ -36,470 +36,472 @@
       <Extensions>
         <desktop4:Extension Category="windows.fileExplorerContextMenus">
           <desktop4:FileExplorerContextMenus>
+            <!-- Prefix Verb Ids with "C" to work around an OS bug preventing the shell menu from appearing in the
+                 context menu on some machines without it. There is no reason found why it works, but it really does -->
             <desktop5:ItemType Type="Directory">
-              <desktop5:Verb Id="MediaInfo" Clsid="20669675-b281-4c4f-94fb-cb6fd3995545" />
+              <desktop5:Verb Id="CMediaInfo" Clsid="20669675-b281-4c4f-94fb-cb6fd3995545" />
             </desktop5:ItemType>
             <desktop5:ItemType Type=".264">
-              <desktop5:Verb Id="MediaInfo" Clsid="20669675-b281-4c4f-94fb-cb6fd3995545" />
+              <desktop5:Verb Id="CMediaInfo" Clsid="20669675-b281-4c4f-94fb-cb6fd3995545" />
             </desktop5:ItemType>
             <desktop5:ItemType Type=".3g2">
-              <desktop5:Verb Id="MediaInfo" Clsid="20669675-b281-4c4f-94fb-cb6fd3995545" />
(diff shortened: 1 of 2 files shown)


JeromeMartinez:
[Snapshot](https://mediaarea.net/download/snapshots/binary/mediainfo-gui/20241216-2/MediaInfo_GUI_24.12.20241216_Windows.exe) @cjee21 @sevastopolcat

cjee21:
@JeromeMartinez Still works fine on my primary PC (Windows 11 24H2 26100.2605, AMD64, 10 apps with packaged context menu shell extension installed).

sevastopolcat:
Yes, it works! The MediaInfo item is in the context menu of media files and folders in Explorer and Total Commander.
Windows 11 26100.2605

JeromeMartinez:
> Yes, it works! The MediaInfo item is in the context menu of media files and folders in Explorer and Total Commander.

So it was not working with "M" as start of the string but "C" is working, that is crazy... But if it is the way, we do that...

Source/WindowsSparsePackage/MSIX/AppxManifest.xml
```diff
@@ -37,469 +37,469 @@
         <desktop4:Extension Category="windows.fileExplorerContextMenus">
           <desktop4:FileExplorerContextMenus>
             <desktop5:ItemType Type="Directory">
```
JeromeMartinez, line 39: Add (in both AppxManifest.xml) something like:
```
            <!-- Prefix with "C" to work around an OS bug preventing the shell menu from appearing in the
                 context menu on some machines without it. There is no reason found why it works, but it really does -->
```

(it will be so easy in the future to remove it for "simplification" if there is no comment)

JeromeMartinez (changes requested):

cjee21:
> So it was not working with "M" as start of the string but "C" is working, that is crazy... But if it is the way, we do that...

Crazy indeed and the fact that it only happens on certain systems.

At least it is not as bad as some Android bugs that cause files to disappear overnight due to special chars in folder names or download of same filename overwriting the previous file silently even after previous has been moved out of downloads folder.

JeromeMartinez (approved):

rlaphoenix:
> [Snapshot](https://mediaarea.net/download/snapshots/binary/mediainfo-gui/20241216-2/MediaInfo_GUI_24.12.20241216_Windows.exe) @cjee21 @sevastopolcat

This snapshot is not resolving this issue on Directory Opus, which does not have the option to use the newer Windows 11-style menu. I also don't know if they use the old-style menu as-is or if its some sort of semi-custom menu.

image
image
image

JeromeMartinez:
> Directory Opus, which does not have the option to use the newer Windows 11-style menu.

If so, how the Opus developer can have it on his machine, see [his own comment](https://github.com/MediaArea/MediaInfo/issues/994#issuecomment-2562449852).
You can read the all thread from the linked comment, it is more complicated than that (and more or less random).

PR #1004, opened 2024-12-29
Update AppIcon for Android

g-maxime:
- white background (more contrast)
- monochrome icon for API 33+ adaptives colors

Before:

DrawerOld
HomeOld

After:

DrawerNew
HomeNew

PR #1010, opened 2025-01-13
Android GUI: Fix square adaptive icon

g-maxime:


JeromeMartinez:
@g-maxime as it was working on some devices with the previous commit, wouldn't it worth it to keep the monochrome element in the v26 file? Or does it do any conformance issue?

JeromeMartinez:
[Snapshot](https://mediaarea.net/download/snapshots/binary/mediainfo-gui/20250113-2/MediaInfo_GUI_24.12.20250113_Android.apk)

JeromeMartinez:
> The [documentation](https://developer.android.com/develop/ui/views/launch/icon_design_adaptive) seems to allow it, but probably useless.

[Seems confirmed](https://github.com/MediaArea/MediaInfo/pull/1004), it works everywhere so I merge this fix as is.

PR #1011, opened 2025-01-14
ShellExt: Handle .url files

cjee21:
Only works on new Windows 11 context menu. Windows File Explorer's classic context menu does not show any shell extensions for .url files.

[URL files for testing.zip](https://github.com/user-attachments/files/18409714/URL.files.for.testing.zip)

### DIFF
--- a/Source/WindowsShellExtension/dllmain.cpp
+++ b/Source/WindowsShellExtension/dllmain.cpp
@@ -33,9 +33,9 @@ BOOL APIENTRY DllMain(_In_ HMODULE hModule, _In_ DWORD  ul_reason_for_call, _In_
 }
 
 namespace {
+
     // Extracted from
     // https://source.chromium.org/chromium/chromium/src/+/main:base/command_line.cc;l=109-159
-
     std::wstring QuoteForCommandLineArg(_In_ const std::wstring& arg) {
         // We follow the quoting rules of CommandLineToArgvW.
         // http://msdn.microsoft.com/en-us/library/17w5ykft.aspx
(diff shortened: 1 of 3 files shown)


cjee21:
Note: #1002 needs to be rebased after this again.

Source/WindowsShellExtension/dllmain.cpp
```diff
@@ -185,6 +185,63 @@ namespace {
...
+        std::ifstream file(filePath, std::ios::binary);
+
+        // Check if file opened successfully
+        if (!file.is_open())
+            return false;
+
+        // Read the entire file content
+        std::string content;
+        file.seekg(0, std::ios::end);
+        content.resize(file.tellg());
```
JeromeMartinez, line 205: There should be a security check that the file is not too big, e.g. leaving if file size is > 4096 bytes (a modern storage block size).

JeromeMartinez (changes requested):
Thank you!
One small addition please:

cjee21:
Does not work on Qt GUI as expected. Need some changes in Qt code to handle URL as command line parameter.

cjee21:
Does not work on Qt GUI as expected. Need some changes in Qt code to handle URL as command line parameter.

It is because of:
https://github.com/MediaArea/MediaInfo/blob/ed69a31ac0cf1b48e00f85adc976508edcbfcb81/Source/GUI/Qt/mainwindow.cpp#L433
Commenting this line out makes it work for urls.

JeromeMartinez:
[Snapshot](https://mediaarea.net/download/snapshots/binary/mediainfo-gui/20250114-2/MediaInfo_GUI_24.12.20250114_Windows.exe).

JeromeMartinez (approved):

JeromeMartinez:
@cjee21 it works great, thanks!

PR #1026, opened 2025-02-12
Migrate MSVC checks from AppVeyor to GitHub Actions + Add check for MediaInfo-Qt

cjee21:
The check for MediaInfo-Qt will only run when there are changes to Qt related files. It is only for Windows x64 and takes less than 10 minutes.


JeromeMartinez:
> The check for MediaInfo-Qt will only run when there are changes to Qt related files.

I was not aware [it was possible](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore) :).

> It is only for Windows x64 and takes less than 10 minutes.

It is fair, so fine.

PR #1028, opened 2025-02-13
Add Ubuntu for MediaInfo-Qt Checks

cjee21:
+ enable colour in msbuild output and use new format for flags
+ prevent 'unreferenced formal parameter' warning due to a work-in-progress part


cjee21:
Thought of checking how newly added graph view works on Ubuntu but looks like MediaInfoLib is not built with graph support:

Screenshot from 2025-02-13 15-06-56

JeromeMartinez:
> on Ubuntu but looks like MediaInfoLib is not built with graph support

@g-maxime how difficult is it to add graphviz on Ubuntu?

PR #1031, opened 2025-02-15
Improve trace feature in VCL GUI and implement it in Qt GUI

cjee21:
- Windows GUI: Improve trace feature
  - Enable switching back to 'Trace level - 0' after enabling higher levels without restarting MediaInfo.
  - Use newer 'Trace_Level' instead of 'Details'.
  - Rename 'Details' to 'Trace level'.
- Qt GUI: Add trace feature
  - Implement trace feature.
  - Only toggle between level 0 and 1.
  - Screenshot:
    Screenshot 2025-02-15 165905

### DIFF
--- a/Source/GUI/Qt/mainwindow.cpp
+++ b/Source/GUI/Qt/mainwindow.cpp
@@ -1231,6 +1231,12 @@ void MainWindow::on_actionFull_Parse_toggled(bool checked)
     refreshDisplay();
 }
 
+void MainWindow::on_actionTrace_toggled(bool checked)
+{
+    C->MI->Option_Static(__T("Trace_Level"), checked?__T("1"):__T("0"));
+    ui->menuView->actions().at(VIEW_TEXT)->trigger();
+}
+
 void MainWindow::on_actionClose_All_triggered()
 {
     C->Menu_File_Open_Files_Begin(true);
--- a/Source/GUI/Qt/mainwindow.h
+++ b/Source/GUI/Qt/mainwindow.h
@@ -88,6 +88,7 @@ private slots:
     void on_actionClose_All_triggered();
     void on_actionAdvanced_Mode_toggled(bool);
     void on_actionFull_Parse_toggled(bool);
+    void on_actionTrace_toggled(bool);
     void on_actionExport_triggered();
     void on_actionPreferences_triggered();
     void on_actionKnown_parameters_triggered();
--- a/Source/GUI/Qt/mainwindow.ui
+++ b/Source/GUI/Qt/mainwindow.ui
@@ -77,6 +77,7 @@
     <addaction name="actionAdapt_columns_to_content"/>
     <addaction name="actionAdvanced_Mode"/>
     <addaction name="actionFull_Parse"/>
+    <addaction name="actionTrace"/>
    </widget>
    <widget class="QMenu" name="menuFile">
     <property name="title">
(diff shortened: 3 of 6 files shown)


JeromeMartinez:
> I think the Qt version is good enough to replace the UWP and wxWidgets versions now 😉

It is nice to have many updates on the Qt version but moving from on toolkit to another is not something I want to do without knowing what I want at long term for not changing user habits too often.
On Linux: the Qt version seems a lot better now than the Wx version and Qt, including QtWebEngine, is well integrated in distros, it may worth it, but I am reluctant to force users to pull the big QtWebEngine module for a feature they don't use (not everyone needs the graphs), so I think I want to make QtWebEngine dependence optional (I may like also the same for graphviz) but not sure yet, I need to try few things for seeing what is doable there before making the Qt version the preferred choice.
On Windows: still not convinced to move away from VCL, on one hand I would love to leave the non open source VCL and related dev environment, on the other side not having integration of the web browser as VCL does (I don't get why [Qt says that that there is no web browser on Windows](https://doc.qt.io/qt-6/qtwebview-index.html) while Embarcadero does it) is a big pain (so many MB for just that... A workaround may be dynamic loading on a plugin containing not only graphviz but also QtWebEngine, but it doesn't resolve the handling of just a basic HTML view, or did I miss something?), so if I move from UWP it would be for the "final" choice (VCL or Qt) and this choice is not yet made.
On macOS: for the free version I may prefer to switch from Wx to the native toolkit (we have it already...) having a kind of incentive for purchasing/donating, when I have time for doing implementation of some stuf + QA.

PS: whatever are the decisions, it is very appreciated to have the updates on the Qt version, it is a back up in case Embarcadero stops to accept us using their toolkit.

cjee21:
> A workaround may be dynamic loading on a plugin containing not only graphviz but also QtWebEngine, but it doesn't resolve the handling of just a basic HTML view, or did I miss something?

Without Qt WebEngine, the HTML view can be displayed but with no borders or proper layout.

Another way is to use Edge WebView2 in Qt. It is possible. I just tried some AI-generated code in an empty Qt project and unexpectedly, it worked right away!

Screenshot 2025-02-16 232907

cjee21:
> I may like also the same for graphviz

graphviz is optional as at now on Linux. It will only work if graphviz package from distro is installed by user else will get a message from MediaInfoLib about not installed.

JeromeMartinez:
> graphviz is optional as at now on Linux. It will only work if graphviz package from distro is installed by user else will get a message from MediaInfoLib about not installed.

great. In practice I didn't check that but I will double check before I switch.

> Another way is to use Edge WebView2 in Qt. It is possible. I just tried some AI-generated code in an empty Qt project and unexpectedly, it worked right away!

Well... Let's give it a try?
I am curious about the size of the resulting binaries compared to VCL (I still have concerns about the Qt version binary size :( )

cjee21:
> great. In practice I didn't check that but I will double check before I switch.

btw PR1029 is safe to merge. It only fixes path of finding graph template HTML. Does not affect compilation or running.

> Well... Let's give it a try? I am curious about the size of the resulting binaries compared to VCL (I still have concerns about the Qt version binary size :( )

I am using dynamic Qt libraries pre-built by Qt though since I did not build Qt. The size may be larger than static since all unused things are also included if in the same dll.

JeromeMartinez:
> I am using dynamic Qt libraries pre-built by Qt though since I did not build Qt. The size may be larger than static since all unused things are also included if in the same dll.

@g-maxime suffers but uses to manage Qt static builds on Windows when there is no QtWebEngine, we'll reuse this knowledge.

PR #1032, opened 2025-02-15
Qt GUI: Implement translations + monospaced font customization

cjee21:
Translations:
Screenshot 2025-02-16 141005
| English | French |
| - | - |
| Screenshot 2025-02-16 140914 | Screenshot 2025-02-16 140900 |
| Screenshot 2025-02-16 140950 | Screenshot 2025-02-16 140935 |

Monospaced font customization (for sheet/text view):
Screenshot 2025-02-16 163005
Screenshot 2025-02-16 163014


cjee21:
After this, others can use Qt Linguist to translate the UI strings.

cjee21:
By the way, CJK languages have misaligned text output in MediaInfoLib.
Is it a known behaviour? @JeromeMartinez

JeromeMartinez (changes requested):
Thank you for the proposal but

> After this, others can use Qt Linguist to translate the UI strings.

No, because the "upstream" is the CSV files, and the Qt translation files need to be updated from theses CSV files.
Having a "fork" in the translation files will be a burden at long term.

Mandatory thing is, as it is done for the macOS translations, an automated script that permits us to easily update the Qt translation files from time to time.
Optionally, if you want that the users are able to use Qt Linguist, an automated script that permits use to easily update the CSV files when there is a Qt files update.

JeromeMartinez:
> By the way, CJK languages have misaligned text output in MediaInfoLib.
Is it a known behaviour? @JeromeMartinez

No, it would deserve a patch.

cjee21:
> Mandatory thing is, as it is done for the macOS translations, an automated script that permits us to easily update the Qt translation files from time to time.

I did use a script to update the Qt translation files from the csv for this PR. Some strings don't match the VCL GUI so they are not translated.

cjee21:
In case it is not clear, the Qt translation is only for the UI part. MediaInfoLib still uses the csv files directly.

JeromeMartinez:
> I did use a script to update the Qt translation files from the csv for this PR.

I don't see the script in the PR.

> Some strings don't match the VCL GUI so they are not translated.

I don't understand.
Anyway, I need to be able to create all the files (ts & qm) from the CSV files. If there is new string not in the VCL, it needs to be added to the CSV files because it would be deleted by the script (I want to be able to create the output files fully, so as if they are not there) + The plan is to have the same features everywhere and the features you added in the Qt version will be at the minimum in the macOS version, and maybe in the VCL version if I keep it longer.

cjee21:
The .ts files are generated from the Qt project/source files using Qt's lupdate CLI tool. They contain all the translatable strings in the Qt GUI. The translations can then be added using Qt Linguist or with a script. The Qt GUI uses different strings for the menus etc so the script cannot match everything properly to the csv. The script is not perfect also. The .qm are generated from .ts using Qt's lrelease CLI tool.

cjee21:
I see no easy way to reliably copy translations from the csv to ts due to the different way Qt handles string identification.

JeromeMartinez:
> The .ts files are generated from the Qt project/source files using Qt's lupdate CLI tool.  [...] The .qm are generated from .ts using Qt's lrelease CLI tool.

Please provide something similar to https://github.com/MediaArea/MediaInfo/tree/master/Source/GUI/Cocoa/_i18n i.e. a documentation and/or a batch for having a quick way to do an update.

>  The Qt GUI uses different strings for the menus etc so the script cannot match everything properly to the csv. The script is not perfect also.

Even if not perfect, but I guess that we could use the English sentences as a "key" for fetching the translation from the other files (English sentences --> CSV first column --> translations), and maybe not creating the files from scratch but being able to update them by catching the English sentence ("source" element) in the Qt files and filing the "translation" element.Creating new lines in the CSV when you have new sentences.

I don't want to have a fork in the translations and/or having to manually fill different places for the same translation.

cjee21:
Okay, fully automated now. I changed some strings in Qt GUI to match as much as possible with existing translation keys. There are still some strings left untranslated as they are no equivalent in the existing translations. I will leave that for others to translate in future. Just need to add those entries in the CSV and run the update script. Existing VCL GUI also has untranslated strings anyway.

JeromeMartinez:
> I don't think I have ever seen order like yours even when manually translating with Linguist.

I just run your .cmd, so weird!
Anyway, not a big deal, and we'll see how we'll manage that when we do the updates.

JeromeMartinez (approved):

cjee21:
> I just run your .cmd, so weird! Anyway, not a big deal,

I just tried again.... On this branch, I deleted everything in Translations folder then run the .cmd.... I got perfect match with this branch, GitHub Desktop shows nothing to commit.

I suspect it may be due to the xml parsing library in Python may be different somehow? Yours seems to be reordering everything in alphabetical order.

JeromeMartinez:
> I suspect it may be due to the xml parsing library in Python may be different somehow? 

Updated my old Python 3 and I have the same thing as you now.

cjee21:
> > By the way, CJK languages have misaligned text output in MediaInfoLib.
> > Is it a known behaviour? @JeromeMartinez
> 
> No, it would deserve a patch.

There is a possible way...

Screenshot 2025-02-18 175551
Screenshot 2025-02-18 175605
Screenshot 2025-02-18 175621

This requires changes to both MediaInfoLib and Qt GUI code.
For Qt GUI, Japanese is imperfect alignment, Korean and Chinese is perfect. For exported txt open in Notepad, Korean and Chinese is perfect, Japanese is misaligned. For VCL GUI, still misaligned.

Regarding Arabic is another different matter. This one is a total mess on both GUIs although they have different kind of mess.

PR #1033, opened 2025-02-17
Qt GUI: Use Edge WebView2 on Windows

cjee21:
It works! :)

I don't know where to put the WebView2 SDK so I just put it in a 'ThirdParty' folder.

The MSIX size after translations and this PR is now around 32MiB. This is with Qt DLLs plus Graph and FFmpeg plugins. May be able to reduce further by removing some unused DLLs (if any remain) or by static linking.

Screenshot 2025-02-17 192335

Screenshot 2025-02-17 192437

### DIFF
--- a/.github/workflows/MediaInfo-Qt_Checks.yml
+++ b/.github/workflows/MediaInfo-Qt_Checks.yml
@@ -3,12 +3,14 @@ name: MediaInfo-Qt Checks
 on:
   push:
     paths:
+      - '.github/workflows/MediaInfo-Qt_Checks.yml'
       - 'Project/QMake/GUI/**'
       - 'Source/GUI/Qt/**'
       - 'Source/Resource/**'
 
   pull_request:
     paths:
+      - '.github/workflows/MediaInfo-Qt_Checks.yml'
       - 'Project/QMake/GUI/**'
       - 'Source/GUI/Qt/**'
       - 'Source/Resource/**'
@@ -19,7 +21,7 @@ jobs:
     runs-on: windows-latest
     strategy:
       matrix:
-        architecture: [ x64 ]
+        architecture: [ x64, ARM64 ]
       fail-fast: false
     steps:
     - name: Checkout zlib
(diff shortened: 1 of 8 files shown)


cjee21:
Full ARM64 version of MediaInfo Qt now builds successfully as well. :)

JeromeMartinez (changes requested):
> It works! :)

Yeah!

> I don't know where to put the WebView2 SDK so I just put it in a 'ThirdParty' folder.

Here should be as much as possible the source, no SDK binaries but maye here a documentation about how to install it.
Then @g-maxime will put them in https://github.com/MediaArea/MediaArea-Utils-Binaries/tree/master/Windows

> The MSIX size after translations and this PR is now around 32MiB. This is with Qt DLLs plus Graph and FFmpeg plugins. May be able to reduce further by removing some unused DLLs (if any remain) or by static linking.

Plugins take ~16 MiB, that means 16 MiB for the main product, relatively small compared to nowadays packages but still 3x more than with VCL. @g-maxime will try a static build while adding the publication of Qt Windows version similar to what we do with other tools we have (but in ~1 month, other priorities).
Then if we start with the Microsoft Store, such size is not a problem anyway (people using the store don't really care of 10 or 30 MiB :) ).

> Full ARM64 version of MediaInfo Qt now builds successfully as well. :)

Great! We'll need the same package as the VCL one so we lose nothing compared to the VCL, even now, because if we release as an alternative of the VCL and/or Microsoft Store, I want that we test everything with that so we already have enough users doing tests before we replace the VCL version.

JeromeMartinez:
Also: there are binary files in the language related directory, I need to know how to recreate them.

cjee21:
> Also: there are binary files in the language related directory, I need to know how to recreate them.

You mean the .qm in the other PR? All are generated from the .cmd as mentioned in the README there.

JeromeMartinez:
> You mean the .qm in the other PR? All are generated from the .cmd as mentioned in the README there.

Mistake (they are in `microsoft.web.webview2.1.0.3065.39/tools/wv2winrt/*/System.CommandLine.resources.dll` so part of the third party which needs to be removed from here), I deleted the message and will delete this one.

cjee21:
Looks like Qt ARM64 builds successfully on GitHub Actions as well. I don't know why the compiler output is empty compared to x64 one though.

> Here should be as much as possible the source, no SDK binaries but maye here a documentation about how to install it.
Then @g-maxime will put them in https://github.com/MediaArea/MediaArea-Utils-Binaries/tree/master/Windows

I'll wait till the SDK are uploaded somewhere then. Need to edit some paths once I know the exact location and test to make sure all still builds. Also need to rebase and make a small edit due to conflict with the other PR anyway.

I do not know the best way to install this SDK. I cannot find it in Visual Studio installer and all Microsoft's guides and samples are using Visual Studio project and installs the SDK via nuget like what is done for winrt/wil on MediaInfo shell extension. I just downloaded the SDK nuget package and extracted the contents for this PR for now.

JeromeMartinez:
> Need to edit some paths once I know the exact location and test to make sure all still builds.  [...] installs the SDK via nuget like

Isn't there a default install dir via nuget?

> I'll wait till the SDK are uploaded somewhere then.

If I understand well, it is "just" a nuget additional stuff, so no upload, it is just added during the install of the compiler, so you can remove the files from here and use the most common patch for this SDK.

Or do I miss something?

cjee21:
> Or do I miss something?

I downloaded Nuget package manually from the site 😅 

I'll see if there is a way to install system wide. For Visual Studio projects like the shell extension, Nuget packages are downloaded to the solution directory on build.

cjee21:
OK. CI can build means it works. No more SDK in here. :)

JeromeMartinez (approved):
A lot better, thank you!

PR #1035, opened 2025-02-18
Fix text view alignment for CJK languages

cjee21:
VCL GUI:
- Select suitable fonts depending on language so that together with https://github.com/MediaArea/MediaInfoLib/pull/2201, proper alignment can be achieved. However, Japanese and Chinese are still slightly misaligned.

Qt GUI:
- Set more preferred fonts so that Qt fallsback to the correct fonts for characters not in the primary font so that together with https://github.com/MediaArea/MediaInfoLib/pull/2201, perfect alignment can be achieved. All CJK languages now have perfect alignment.
- If user sets their own font in Preferences, that will have priority and Qt will only use fonts in preferred list only for characters not available in their chosen font.

Language files:
- Fix Chinese (CN) column size not updated to be coherent with other languages, causing cropped text.


cjee21:
Before: https://github.com/MediaArea/MediaInfo/pull/1032#issuecomment-2661274315

JeromeMartinez:
Thank you.

> However, Japanese and Chinese are still slightly misaligned

Wouldn't it be an issue with the choice of a default font? I guess that some fonts are fine, but it needs some tests (and implementation of font selection in the VCL version).
But fair enough as is, so merged and I'll see later for the font selection.

cjee21:
> Wouldn't it be an issue with the choice of a default font?

I used same font on VCL, Qt and Notepad... Only the VCL is still misaligned. No idea.

Arabic looks completely messed up on both GUIs and in different ways on both.

PR #1037, opened 2025-02-21
Fix MediaInfo-Qt ARM64 CI

cjee21:
Turns out the CI did not actually build the ARM64 version of Qt GUI. Qt uses batch file for ARM64 `qmake` which links to x64 exe `qmake`. So a `call` is needed infront of qmake call for ARM64 else the command(s) in the following line(s) will not be executed.


JeromeMartinez:
Good catch!

PR #1038, opened 2025-02-22
Qt GUI: Add Conformance glossary button

cjee21:
Just remembered there is this button. Now Qt GUI also has it though the webpage that the button leads to appears to have nothing.

Screenshot 2025-02-22 175734


cjee21:
Ah forgot that this is cross-platform so need something for ShellExecute on Linux...

JeromeMartinez:
Checking the changes, do Qt translation files change so much each time you add a line of code? It is creating a strong dependency of commits and also a big Git log for any small change...
I was already not happy with the VCL translation files (too big, too many change for any translation) but now it becomes even bigger (translation files update even with translation update) :(.

JeromeMartinez:
> appears to have nothing.

It is currently broken, I need to fix that ASAP.

cjee21:
> Checking the changes, do Qt translation files change so much each time you add a line of code? It is creating a strong dependency of commits and also a big Git log for any small change... I was already not happy with the VCL translation files (too big, too many change for any translation) but now it becomes even bigger (translation files update even with translation update) :(.

Other then the new string for the button, the lines of code changes but even if those are wrong the app still works. I think that is for Qt Linguist so that it can show where the string comes from for translators to refer.

Maybe we remove translation files from git and generate them on every build?

JeromeMartinez:
> Maybe we remove translation files from git and generate them on every build?

I think so. Actually also for the big all in one file (I actually use it only as an intermediate file) but the code for creating it is awful, on my todo list. And same for the macOS translation files. In the meantime, having a script for creating the the Qt translation files may be better and a first step in the expected direction for the source code.

cjee21:
But the generation script will have to work cross-platform in that case... 🤔

JeromeMartinez:
> But the generation script will have to work cross-platform in that case... 🤔

For the first langage files (the csv ones), I use... C++ as crossplatform script :).

cjee21:
Rebased

PR #1039, opened 2025-02-22
Qt GUI: Remove translations files from Git and generate them before build instead + remove WebView2 SDK version dependency

cjee21:


cjee21:
> Is it difficult to avoid to rely on a very specific version of Qt?

There should be no problem buulding it with different versions of Qt though recently I only tested with latest. Just that need at least 6.5 for Windows dark mode support and 6.8 for official MSVC2022 (the older one with MSVC2019 is actually compatible with MSVC2022 too).

For the CI, the reason for the version hardcode is due to the path to some Qt executables contain the version.

cjee21:
I'll modify CI with a variable for version so that easier to change.

cjee21:
Removed version dependencies for Qt and WebView2. Now only one env var in CI for Qt version that is set to latest version of Qt 6 which results in using the pre-release 6.9.0 currently.

JeromeMartinez:
> Removed version dependencies for Qt and WebView2.

Thank you.

>  Now only one env var in CI for Qt version that is set to latest version of Qt 6 which results in using the pre-release 6.9.0 currently.

Totally fine (and better, actually, no need of any manual update and we anticipate issues).

PR #1041, opened 2025-02-28
configure.ac: Use proper $PKG_CONFIG variable.

navi-desu:
Necessary for cross-compilation to work, where pkg-config will be set to ${CHOST}-pkg-config.


JeromeMartinez:
Thanks.

PR #1045, opened 2025-03-02
Qt GUI: Improvements

cjee21:
- Add ability to resolve Windows `.lnk` shortcuts
- Make sheet view vertical split resizable
- Make contents of Preferences window auto resize when window is manually resized to make use of available window space.
- Add missing refresh after changing current view after Preferences are changed. It may appear to work without refresh after translations commit added a refresh in the language change event but that should not be relied upon for this purpose.
- Ensure created C++ objects are deleted
- Fix type comparison warning
- Remove outdated Qt MSVC2022 project that no longer works. Instead, make changes to Qt project file and add a cmd script that can generate a working MSVC2022 project when needed.


JeromeMartinez:
thanks.

> Remove outdated Qt MSVC2022 project that no longer works.

True, so let's do this way.

cjee21:
A test build of MediaInfo Qt GUI in MSIX is available here: https://github.com/cjee21/MediaInfo/actions/runs/14009285448

Since it is unsigned, it can be installed on Windows 11 with the following command in an Administrator PowerShell:
```powershell
Add-AppxPackage -Path MediaInfo_Qt_Windows_x64.msix -AllowUnsigned
```

PR #1054, opened 2025-03-15
Update .gitignore for MSVC

cjee21:
Tidy/clean-up `.gitignore` and make it consistent with MediaInfoLib for the MSVC section.

Nuget section is now an exact copy from https://github.com/github/gitignore/blob/main/VisualStudio.gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -40,34 +40,22 @@ Win32
 x64
 My Project
 
-# Visual Studio - Roslyn cache directory
+# Visual Studio 2015+ cache/options directory
 .vs/
 
-# NuGet -------------------------------------------------------------------------------------------
-
-# Ignore NuGet Packages
+# NuGet Packages
 *.nupkg
-
+# NuGet Symbol Packages
+*.snupkg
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -40,35 +40,23 @@ Win32
...
 #!**/[Pp]ackages/repositories.config
-
 # NuGet v3's project.json files produces more ignorable files
 *.nuget.props
 *.nuget.targets
 
-# Ignore other intermediate files that NuGet might create. project.lock.json is used in conjunction
-# with project.json (NuGet v3); project.assets.json is used in conjunction with the PackageReference
-# format (NuGet v4 and .NET Core).
-project.lock.json
```
JeromeMartinez, line 67: Theses files are in https://github.com/github/gitignore/blob/main/VisualStudio.gitignore#L61, what do I miss?

cjee21: They are only used for .NET projects that in certain configurations. I've not seen those files in MediaInfo.

PR #1060, opened 2025-03-21
Preparing v25.03

g-maxime:


History_GUI.txt
```diff
@@ -12,6 +12,39 @@ Known bugs
 - Languages (other than english and French) : not all words are translated, I need translators!
 - Others? https://sourceforge.net/p/mediainfo/_list/tickets
 
+Version 25.03, 2025-03-21
+-------------
++ Italian treanslation update
```
cjee21, line 17: Typo

JeromeMartinez: good catch, we could fix here but I think we won't relaunch the release build, it is not worth it to delay the release for that.

PR #1064, opened 2025-03-23
Qt GUI: Fixes for Windows 10

cjee21:
- Enable name display on Windows 10 tile
  - Before and after:
   Screenshot 2025-03-23 163641 Screenshot 2025-03-23 164458
- Fix font selection on Windows when Cascadia Mono in unavailable

Tested that everything is fine on Windows 10 and 11. It also has the correct look/design/theme depending on Windows version.

A test build of MediaInfo Qt GUI in MSIX is available here: https://github.com/cjee21/MediaInfo/actions/runs/14017355671

Since it is unsigned, it can be installed on Windows 11 with the following command in an Administrator PowerShell:
```powershell
Add-AppxPackage -Path MediaInfo_Qt_Windows_x64.msix -AllowUnsigned
```

### DIFF
--- a/Source/GUI/Qt/mainwindow.cpp
+++ b/Source/GUI/Qt/mainwindow.cpp
@@ -538,9 +538,9 @@ void MainWindow::refreshDisplay() {
     QFont font;
     if (!settings->value("monoFont", "").toString().isEmpty()) {
         font.fromString(settings->value("monoFont", "").toString());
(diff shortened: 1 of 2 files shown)


cjee21:
One more thing that Qt is better than VCL, when tags contain emoji, Qt has colour emoji in text view.

JeromeMartinez:
> Enable name display on Windows 10 tile 

Weird, it is the default with the VCL version.

PR #1081, opened 2025-04-05
Qt GUI: Allows to build without HTML-based views

g-maxime:
QtWebEngine isn't packaged for i386 and some other archs by Ubuntu.

+ Fix Qt5


cjee21:
FYI there is also display layout issue with some older Qt versions due to change to fully 
qualified enumerations in .ui files for newer Qt: https://bugreports.qt.io/browse/QTBUG-126966. This affects layout of Preferences window when using certain Qt versions.

Another thing is in Qt 6.9, the highlight colour in sheet view changed to follow Windows accent colour which may not look as nice as before.

Screenshot 2025-04-06 203944
Screenshot 2025-04-06 203622

PR #1082, opened 2025-04-06
Qt GUI: Fixes for Qt<6.7 and translations generation

cjee21:
Fix Preferences window layout on older Qt versions used by some Linux distros due to change change to fully
qualified enumerations in `.ui` files for newer Qt (https://bugreports.qt.io/browse/QTBUG-126966). This is done by replacing the affected enumerations in `.ui` files when building with older Qt versions.

Some other fixes mainly for translations generation after the changes made in #1081.

Fixed layout on Ubuntu 24.04.2 using repo provided Qt 6.4.2:
Screenshot from 2025-04-06 22-15-28

### DIFF
--- a/Release/Release_GUI_Qt_Windows_x64_MSIX.cmd
+++ b/Release/Release_GUI_Qt_Windows_x64_MSIX.cmd
@@ -1,7 +1,7 @@
 :: config
 ::  example:
-::   set QT_PATH=D:\Qt\6.8.2\msvc2022_64
-::   set QT_TOOLS_PATH=D:\Qt\Tools
+::   set QT_PATH=C:\Qt\6.9.0\msvc2022_64
+::   set QT_TOOLS_PATH=C:\Qt\Tools
 ::   set FFMPEG_EXE=%~dp0\..\..\MediaInfo-FFmpeg-Plugin\ffmpeg.exe
 ::   set CERT_PATH=D:\sign_cert.pfx
 ::   set CERT_PASS=K1wdqYSDk0locw6HSSjT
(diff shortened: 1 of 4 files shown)


cjee21:
For Microsoft Store, it should be deploy-able as it is. Just need make minor changes to the manifest xml and signing part of build script. For ARM64, need ARM64 version of graphviz and FFmpeg if want to include that.

For Linux, installing graphviz package from repo is sufficient to have graph view working. Just that the graph template HTML needs to be included somewhere such as in the GUI package.

cjee21:
Added ability to use Qt GUI installed with MSIX from command line. For example `mediainfo-gui.exe <filepath or URL>`.

Release/Release_GUI_Qt_Windows_x64_MSIX.cmd
```diff
@@ -1,7 +1,7 @@
 :: config
 ::  example:
-::   set QT_PATH=D:\Qt\6.8.2\msvc2022_64
-::   set QT_TOOLS_PATH=D:\Qt\Tools
+::   set QT_PATH=C:\Qt\6.9.0\msvc2022_64
```
JeromeMartinez, line 3: Isn't it possible to have something more generic for not having to update the version each time we update Qt?

cjee21:
> By the way, for you information we opted for:
> 
>     * Step 1 (on the way): Qt version in Snap/FlatPack/AppImage
> 
>     * Step 2 (soon): Qt version in Windows Store
> 
>     * Step 3 (this year): Qt version in Linux distros
>       We do step by step for not having to manage too many issue reports in one wave.

Nice plan. In my opinion, the Microsoft Store and Linux will benefit most since the current one is very lacking compared to Windows.

>       Qt version by default on Windows is not yet decided.

I think this one way later because Qt has dropped support for out-of-support Windows versions and also there are some things that are only present in VCL version so need to keep that one to prevent complains.

Release/Release_GUI_Qt_Windows_x64_MSIX.cmd
cjee21: This is just how the current build script works at the moment since the path to Qt is not known. That is just an example based on default Qt install path. Before using this script, just set QT_PATH to point to whatever Qt path it is in your specific installation or if it is already in the system PATH, no need to do anything.

I'm just updating the example. This file does not need to be updated when Qt is updated or changed path.

PR #1085, opened 2025-04-10
Windows GUI: Only refresh once ready on startup

### LINKED ISSUE #1084: Files fail to open in HTML view
With latest version, 25.03, on Win 10, video files (.mkv, .mp4) fail to open when the HTML view is selected as default. 
(You must at least open one file) is displayed. 
I can select the HTML from the view drop down or change any settings and it displays correctly. 
I have completely removed everything associated with MediaInfo  and done a fresh install with no change in behavior. 
I also have version 24.xx installed on my laptop which displays correctly when opening the same files.  

This is the GUI version installed from the universal installer and the files are opened from the Explorer extension.

Any help or pointers would be much appreciated!

cjee21:
Fix #1084

### DIFF
--- a/Source/GUI/VCL/GUI_Main.cpp
+++ b/Source/GUI/VCL/GUI_Main.cpp
@@ -286,6 +286,7 @@ __fastcall TMainF::TMainF(TComponent* Owner)
     #endif
 
     //Refresh after opening files
+    StartupReady = true;
     Refresh();
 }
 
@@ -807,6 +808,9 @@ void __fastcall TMainF::Translate()
 //---------------------------------------------------------------------------
 void __fastcall TMainF::Refresh(TTabSheet *Page)
 {
+    if (!StartupReady)
+        return;
+
     if (Page==NULL)
         Page=this->Page->ActivePage;
     size_t FilesCount=I->Count_Get();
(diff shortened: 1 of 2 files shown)


cjee21:
@JeromeMartinez while testing, noticed that https://github.com/MediaArea/MediaInfo/pull/944#issuecomment-2464680617 still happens randomly sometimes with IE engine and HTML set as default with addresses that sometimes vary each time. No idea if it affects production systems or is a result of my testing method to force fallback to IE since I cannot uninstall WebView2.

cjee21:
@JeromeMartinez let's try this. Moved refresh to FormShow which seems to prevent access violations. So far I did cannot reproduce access violations after this change. Maybe using webview in form constructor is too early and form is not yet ready. The boolean flag serves two purpose, one is to prevent refresh from being called multiple times and too early (there are some stuff in GUI_configure that triggers a refresh), two is to ensure the refresh in FormShow is only triggered once during application startup since other things such as changing themes also trigger FormShow.

JeromeMartinez:
> So far I cannot reproduce access violations after this change.

Thank you for fixing my 20 year old buggy code (I try not to read my old code for not becoming too much ashamed). Let's see that! [snapshot](https://mediaarea.net/download/snapshots/branches/binary/mediainfo-gui/20250410_VCL/MediaInfo_GUI_25.03.20250410_Windows.exe)

cjee21:
> Thank you for fixing my 20 year old buggy code (I try not to read my old code for not becoming too much ashamed).

This specific access violation is not due to 20 year old code though. It is probably due to previous changes for accommodating theme switching and WebView2.

> [snapshot](https://mediaarea.net/download/snapshots/branches/binary/mediainfo-gui/20250410_VCL/MediaInfo_GUI_25.03.20250410_Windows.exe)

Tested tens of times with both IE and WebView2 with HTML view set as default and there are no access violations so I consider it fixed.

PR #1087, opened 2025-04-13
Qt GUI: Fix SVG interference in Graph view

cjee21:
Isolate the each SVG in shadow DOMs to prevent them interfering with each other and causing text corruption due to Graphviz Cairo using same IDs each time for glyphs.

Issue: https://github.com/MediaArea/MediaInfoLib/issues/2247
Requires updated graph template for working zoom:

PR #1090, opened 2025-04-15
Qt GUI: Improve loading of large graphs in Graph view

cjee21:
Remove the imperfect Edge WebView2 size limitation workaround (#1087).

Use a temporary file for Graph view HTML to ensure large graphs can load and display properly in both Edge WebView2 and Qt WebEngine.

Tested on Windows + Edge WebView2 and Ubuntu + Qt WebEngine.

Screenshot 2025-04-15 231549
Screenshot from 2025-04-15 23-16-03

### DIFF
--- a/Source/GUI/Qt/graphplugin.cpp
+++ b/Source/GUI/Qt/graphplugin.cpp
@@ -8,7 +8,6 @@
 
 #include "graphplugin.h"
 #include "ZenLib/Ztring.h"
-#include "ZenLib/File.h"
 #include <QApplication>
 #include <QComboBox>
 #include <QDir>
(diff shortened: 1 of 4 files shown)


Source/GUI/Qt/graphplugin.cpp
```diff
@@ -43,6 +43,7 @@ GraphViewWidget::GraphViewWidget(Core *C, QSettings *settings, QWidget *parent)
 
     this->setLayout(layout);
 
+    tempFile.setFileTemplate(QDir::tempPath() + QDir::separator() + QCoreApplication::applicationName() + "_XXXXXX.html");
```
JeromeMartinez, line 46: QTemporaryFile default name not good?

Source/GUI/Qt/graphplugin.cpp
```diff
@@ -82,7 +83,11 @@ QString GraphViewWidget::Generate_Graph_HTML() {
 }
 
 void GraphViewWidget::refresh() {
-    webView->setHtml(Generate_Graph_HTML());
```
JeromeMartinez, line 85: I like the idea of not having a temporary file when we can avoid it, possible to keep the previous style when it the content is not too big?

Source/GUI/Qt/graphplugin.cpp
cjee21: lacks `.html` which then causes WebView to show the source code instead of rendering the HTML.

JeromeMartinez: > which then causes WebView to show the source code instead of rendering the HTML.

Relying on the file extension, ha ha.

> lacks .html

Below something more long term?
"tempFile.setFileTemplate(tempFile.fileTemplate() + ".html");"

Source/GUI/Qt/graphplugin.cpp
cjee21: Done. Used 1MB as limit since the actual limit is uncertain. The file that failed is ~1.7MB.

cjee21: I think safer to use 0.5MB. For Edge WebView2, 1MB is probably safe since others say they found the limit to be around 1.5MB so probably Microsoft uses base64 encoding. However, I checked Qt WebEngine sourcecode and found they are really using percent encoding as their docs said so the worst case should be around 3x size increase?

Source/GUI/Qt/graphplugin.cpp
JeromeMartinez: > "tempFile.setFileTemplate(tempFile.fileTemplate() + ".html");"

ping

cjee21: This works

cjee21:
eh wait, there's a mistake

cjee21:
Re-tested graph view on both Windows and Ubuntu. Still works.

PR #1095, opened 2025-04-18
Windows/Qt GUI: Rename `LegacyStreamDisplay` option + add translation

cjee21:
To better reflect what the option really does and enable others to translate it.


JeromeMartinez:
@cjee21 please rebase.

PR #1109, opened 2025-05-02
Fix crash in mediainfo-gui snap

g-maxime:
Already available on snapcraft.io


cjee21:
Confirmed that opening using file dialog and exporting works on Ubuntu 24.04.2 now.

PR #1118, opened 2025-05-17
build(cli): add cmake support

### LINKED ISSUE #1117: CMake support?
will it possible to add CMake build system support for CLI?

trim21:
close #1117

### DIFF
--- a/.github/workflows/MediaInfo_Checks.yml
+++ b/.github/workflows/MediaInfo_Checks.yml
@@ -142,3 +142,94 @@ jobs:
             Write-Host "::error::Test CLI: MediaInfo CLI output ($actual_output) is not as expected ($expected_output)"
             exit 1
           }
+
+  cmake:
+    env:
+      CC: cl.exe
+      CXX: cl.exe
+    strategy:
+      matrix:
+        include:
+          - architecture: x64
+            runner: windows-latest
+            Format_Profile: Executable / AMD x86-64
+          - architecture: ARM64
+            runner: windows-11-arm
+            Format_Profile: Executable / ARM64
+      fail-fast: false
+    name: CMake windows (${{ matrix.architecture }})
+    runs-on: ${{ matrix.runner }}
+    steps:
+      - name: Checkout zlib
+        uses: actions/checkout@v4
+        with:
+          repository: MediaArea/zlib
+          path: zlib
+      - name: Checkout ZenLib
+        uses: actions/checkout@v4
+        with:
+          repository: MediaArea/ZenLib
+          path: ZenLib
+      - name: Checkout MediaInfoLib
+        uses: actions/checkout@v4
+        with:
+          repository: MediaArea/MediaInfoLib
+          path: MediaInfoLib
+
+      - name: Checkout MediaInfo
+        uses: actions/checkout@v4
+        with:
+          path: MediaInfo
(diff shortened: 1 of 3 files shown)


.github/workflows/MediaInfo_Checks.yml
```diff
@@ -142,3 +145,97 @@ jobs:
...
+        uses: actions/checkout@v4
+        with:
+          repository: MediaArea/ZenLib
+          path: ZenLib
+      - name: Checkout MediaInfoLib
+        uses: actions/checkout@v4
+        with:
+          repository: MediaArea/MediaInfoLib
+          path: MediaInfoLib
+      - name: Checkout MediaArea-Utils for MediaInfo_NoGUI patch
```
cjee21, line 181: No need this since not using MSVC project.

.github/workflows/MediaInfo_Checks.yml
```diff
@@ -142,3 +145,97 @@ jobs:
...
+          cmake -G Ninja -D CMAKE_PREFIX_PATH=./cmake-prefix/ \
+            -D CMAKE_INSTALL_PREFIX=./cmake-prefix/ \
+            -D CMAKE_BUILD_TYPE=Release \
+            -B build \
+            ./MediaInfo/Project/CMake/CLI/
+
+          cmake --build build
+
+
+      - run: ./build/mediainfo.exe --help
```
cjee21, line 235: Test by at least doing mediainfo.exe mediainfo.exe as just doing --help does not test if it can load mediainfo.dll properly since you are using MEDIAINFO_DLL_RUNTIME.

JeromeMartinez: I prefer a static build by default (with option for dll runtime?) if not too complicated.
The test in the CI is currently with testing that the result of e.g. ".\MediaInfo.exe" "--Output=General;%Format_Profile%" "MediaInfo.exe" is the expected one.

trim21: actually I think it won't work, I didn't add dll path to env.

cjee21: I made a local build with this PR and copied the dll but still fail `Unable to load MediaInfo.dll`

cjee21: Okay, I see what is going on now, `zlib.dll` is needed too.

trim21: > I prefer a static build by default (with option for dll runtime?) if not too complicated.


After hours trying, I think it's current impossible to get a fullly static linked executable using cmake with current upstream package config, I'll only test the dynamic linked version.

cjee21: Ya I keep getting zlib linking issues after trying many times too with my version that can build everything in one cmake command.

trim21: I hope this won't be a blocker on merging this PR

JeromeMartinez: > Requires using zlib develop branch

Latest release is not enough?

>  and updating both MediaInfoLib and MediaInfo CMake files.

Please send PRs, so we can prepare that and I'll merge them when we update our zlib.

PR #1120, opened 2025-05-18
Update Android project

g-maxime:
- Update to API 35
- Sets Minimum API to 21 (Android 5.0)
- Allows 16kb pagesize
- Update dependencies

### DIFF
--- a/Source/GUI/Android/app/CMakeLists.txt
+++ b/Source/GUI/Android/app/CMakeLists.txt
@@ -3,6 +3,8 @@
 
 cmake_minimum_required(VERSION 3.4.1)
 
+project(MediaInfo)
+
 # Build Zenlib
 set(ZenLib_DIR ${CMAKE_CURRENT_BINARY_DIR}/ZenLib)
 set(ZenLib_LIBRARY zen)
--- a/Source/GUI/Android/app/build.gradle
+++ b/Source/GUI/Android/app/build.gradle
@@ -1,25 +1,23 @@
 apply plugin: 'com.android.application'
-
+apply plugin: 'com.google.devtools.ksp'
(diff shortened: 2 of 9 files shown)


g-maxime:
Perhaps we could set the 25.05 release as the last version for Android 4.x and merge it after?

cjee21:
> Not a significant bug because 64-bit is always more or same as size_t width on all known CPUs, but easy to be more correct (replace int64u by size_t).

But after that there is `if (ExpectedFileSize!=(int64u)-1)`. Wouldn't it cause this to always be true?

cjee21:
> Not a significant bug because 64-bit is always more or same as size_t width on all known CPUs, but easy to be more correct (replace int64u by size_t).

On second look, you are right. It will be set to max value anyway.

JeromeMartinez:
> On second look, you are right.

It does not prevent me to be interested in fix of such kind of warnings ;-).

cjee21:
It seems one of the MediaInfo GUI dependencies for Android no longer exists:

```
FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':app:dataBindingMergeDependencyArtifactsRelease'.
> Could not resolve all files for configuration ':app:releaseCompileClasspath'.
   > Could not find com.github.YarikSOffice:lingver:1.2.1.
     Required by:
         project :app
```

cjee21:
Since minSDK is increased now means the third-party file picker that's no longer maintained is not needed anymore right?

g-maxime:
> Since minSDK is increased now means the third-party file picker that's no longer maintained is not needed anymore right?

Yes, I'll remove it.

g-maxime:
PR ready

PR #1122, opened 2025-05-21
Windows GUI: Retain legacy shell option

### LINKED ISSUE #994: Windows 11 integration need an option for keeping the old context menu

cjee21:
Resolves #994


cjee21:
This PR is not tested at all.

JeromeMartinez:
@cjee21 FYI I only removed the "Node" part of your patch and it seems that the index issue disappears in that case.

cjee21:
Fixed some copy-paste errors and a quick test, it appears to work.

JeromeMartinez:
> it appears to work.

Confirmed by one of the users having the issue.

PR #1141, opened 2025-07-30
GUI: Some fixes and improvements

cjee21:
- PreRelease
  - Make the PreRelease app build-able on latest C++Builder
  - Fix error checking as `ZtringListListF::Load` returns a `bool`
  - Update languages list
- Translations
  - Sync translations as previous PR did not sync `Languages.csv` which is used for generating Qt translations.
  - Add id translations for newly added GUI strings
- Add .obu file extension
  - For raw AV1 OBU
- Qt GUI
  - Optimize the number of refreshes by removing unnecessary refreshes


cjee21:
The `Abort Opening` button in Qt GUI open progress bar dialog does not actually abort the opening. It just dismisses the dialog while the opening process continues in the background. I don't know if there is a function in MediaInfoLib for aborting.

JeromeMartinez:
> Sync translations as previous PR did not sync Languages.csv which is used for generating Qt translations.

I think I'll remove this file soon because it is redundant, I'll add a script for building it so you add it to your own script.

>  I don't know if there is a function in MediaInfoLib for aborting.

Try `MediaInfoList::Close()` from the other thread, the processing thread should stop not long after the call, but I didn't try that for a long time, no promise that it works.

cjee21:
> > Sync translations as previous PR did not sync Languages.csv which is used for generating Qt translations.
> 
> I think I'll remove this file soon because it is redundant, I'll add a script for building it so you add it to your own script.

Or maybe the Qt translations generation script can be updated to directly use the individual language files.

> > I don't know if there is a function in MediaInfoLib for aborting.
> 
> Try `MediaInfoList::Close()` from the other thread, the processing thread should stop not long after the call, but I didn't try that for a long time, no promise that it works.

`Close()` does stop the parsing but the problem is if closed while it is still parsing, after that it will not open files properly anymore for that session. It appears to read files (I see disk activity) but the progress bar will hang at a certain percentage.

JeromeMartinez:
> It appears to read files (I see disk activity) but the progress bar will hang at a certain percentage.

OK, I'll try to reproduce and fix.

cjee21:
> OK, I'll try to reproduce and fix.

What I used to test:

```diff
diff --git a/Source/GUI/Qt/mainwindow.cpp b/Source/GUI/Qt/mainwindow.cpp
index 2d0136b5..7487b8fe 100644
--- a/Source/GUI/Qt/mainwindow.cpp
+++ b/Source/GUI/Qt/mainwindow.cpp
@@ -508,6 +508,9 @@ void MainWindow::updateProgressBar ()

     if (C->State_Get()==10000 || progressDialog->wasCanceled())
     {
+        if (progressDialog->wasCanceled())
+            C->Menu_File_Open_Files_Begin(true);
+
         progressDialog->hide();
         timer->stop();
```

PR #1143, opened 2025-07-31
Qt GUI: Source translations from individual CSVs

cjee21:
As mentioned at https://github.com/MediaArea/MediaInfo/pull/1141#issuecomment-3140042194, now Qt translations no longer depends on `Languages.csv` so it is safe to be removed.


cjee21:
Also noticed some more redundant and outdated files such as https://github.com/MediaArea/MediaInfo/blob/master/Source/GUI/Cocoa/fr.lproj/lang.csv.

JeromeMartinez:
> Also noticed some more redundant and outdated files such as https://github.com/MediaArea/MediaInfo/blob/master/Source/GUI/Cocoa/fr.lproj/lang.csv.

We need to remove it too.

PR #1146, opened 2025-08-03
Android GUI: Migrate to Material Design 3

cjee21:
Migrate to Material 3 design and enable dynamic theming along with fixing invisible white status bar icons. Also make 'About' page scroll-able when needed (for example landscape mode on small devices or split screen).

### DIFF
--- a/.github/workflows/MediaInfo-Android_Checks.yml
+++ b/.github/workflows/MediaInfo-Android_Checks.yml
@@ -1,0 +1,61 @@
+name: MediaInfo-Android Checks
+
+on:
+  push:
+    paths:
+      - '.github/workflows/MediaInfo-Android_Checks.yml'
+      - 'Source/GUI/Android/**'
+
+  pull_request:
+    paths:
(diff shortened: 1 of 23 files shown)


cjee21:
There are still stuff overlapping with cutouts and system bars with the edge to edge commit.

cjee21:
> There are still stuff overlapping with cutouts and system bars with the edge to edge commit.

Should be okay now. Tested on Android 9, 14, 16:

Screenshot_20250804_212008



Screenshot_20250804_212157

Screenshot_20250804_212226

cjee21:
I noticed there is review on Google Play about font selection. I tested on Xiaomi and Samsung. On Xiaomi, changing system font changes all font in MediaInfo app whereas on Samsung, all font follows the set font except the report itself.
Since I do not see font related stuff in MediaInfo, I think it is dependent on how the OEM implements font selection.

cjee21:
Rebased after PR1162 which already implement scrolling for subscribe page.

JeromeMartinez:
Thank you!

PR #1148, opened 2025-08-04
Android GUI: Update billing API

g-maxime:


cjee21:
FYI I just found out the currently available on Google Play version has a bug. If Preferences page is opened first, the buttons on the Subscribe page will not work.

JeromeMartinez:
>  If Preferences page is opened first, the buttons on the Subscribe page will not work.

Thanks for the report, @g-maxime will try to reproduce & fix.

PR #1161, opened 2025-09-17
Sponsors widgets fixes

g-maxime:
- Set the cursor to a hand when hovering (GUI and installer)
- Fix various scaling problems
- Fix the cropped image in hiDPI mode
- Fix the frame dimensions after maximising

[Snapshot](https://mediaarea.net/download/snapshots/branches/binary/mediainfo-gui/20250917_sponsor/MediaInfo_GUI_25.07.20250917_Windows.exe)

### DIFF
--- a/Source/GUI/VCL/GUI_Main.cpp
+++ b/Source/GUI/VCL/GUI_Main.cpp
@@ -681,6 +681,10 @@ void __fastcall TMainF::FormResize(TObject *Sender)
         Page_System_Sheet->Width = Page_System->ClientWidth - 2;
         Page_System_Sheet->Height = Page_System->ClientHeight - Page_System_Sheet->Top - 2;
     }
+
+    //Footer
+    if (Footer_Sponsor && Footer_Sponsor->Visible)
+        Footer_Sponsor->OnResize(NULL);
 }
 
 //---------------------------------------------------------------------------
@@ -808,6 +812,9 @@ void __fastcall TMainF::Translate()
     M_NewVersion->Caption=(__T(" | ")+Prefs->Translate(__T("NewVersion_Menu"))).c_str();
     M_NewVersion->Visible=Prefs->NewVersion_Display;
 
+    //Footer
+    Footer_Button->Caption=L"Go to conformance errors and warnings glossary page";
+
     //Sponsor
     M_Sponsor->Visible=false;
     if (Prefs->Sponsored && !Prefs->Donated && !Prefs->Translate(__T("SponsorMessage")).empty() && !Prefs->Translate(__T("SponsorMessage")).empty())
@@ -817,9 +824,6 @@ void __fastcall TMainF::Translate()
     }
     if (Footer_Sponsor && Footer_Sponsor->Visible)
         Footer_Sponsor->Translate();
-
-    //Footer
-    Footer_Button->Caption=L"Go to conformance errors and warnings glossary page";
 }
 
 //---------------------------------------------------------------------------
--- a/Source/GUI/VCL/GUI_Sponsor.cpp
+++ b/Source/GUI/VCL/GUI_Sponsor.cpp
@@ -175,30 +175,27 @@ void __fastcall TSponsorFrame::Finalize()
     } catch(...) {}
 }
 
-void __fastcall TSponsorFrame::FormResize(TObject *Sender)
-{   if (!Visible)
+//---------------------------------------------------------------------------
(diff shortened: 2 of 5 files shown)


cjee21:
> If you want to test but already closed the ad, you need to edit C:\Users\%USERNAME%\AppData\Roaming\MediaInfo\Plugin\MediaInfo.cfg and set both (I am not sure exactly about which one is relevant) Install and FirstInstall to the indicated value minus 1000000.

Decrement both `Install` and `SponsorBannerClose`. `FirstInstall` does not matter.

Maybe can still be improved as in some cases at 100%, the ad looks very large compared to the rest of the UI.

100%:
Screenshot 2025-09-18 010754
Screenshot 2025-09-18 010841
Screenshot 2025-09-18 010907
Screenshot 2025-09-18 010927

200%:
Screenshot 2025-09-18 011351

cjee21:
Based on observation (I did not look through the codes) it seems it is scaled full width and the height is capped but it appears the same height in pixels is used for all DPI. The height limit should vary depending on DPI.

JeromeMartinez:
@cjee21 [updated snapshot](https://mediaarea.net/download/snapshots/branches/binary/mediainfo-gui/20250919_sponsor/MediaInfo_GUI_25.07.20250919_Windows.exe) with as height limited at 25% of the Window height.

cjee21:
> @cjee21 [updated snapshot](https://mediaarea.net/download/snapshots/branches/binary/mediainfo-gui/20250919_sponsor/MediaInfo_GUI_25.07.20250919_Windows.exe) with as height limited at 25% of the Window height.

Better. The only thing is when maximised on a high resolution display at 100%, such as on 1440p 27" display, the ad looks big with a height of almost 8 cm (the 2nd image out of the 5 images in the above post). It occupies less than 25% of the window but looks very large compared to the other UI elements so I don't know if it's a good idea.

g-maxime:
[latest snapshot](https://mediaarea.net/download/snapshots/branches/binary/mediainfo-gui/20250924_sponsor/MediaInfo_GUI_25.07.20250924_Windows.exe)

cjee21:
> [latest snapshot](https://mediaarea.net/download/snapshots/branches/binary/mediainfo-gui/20250924_sponsor/MediaInfo_GUI_25.07.20250924_Windows.exe)

Better. Now does not look too large at 100%. Should be good enough. Now ad is kind of small at high DPI. It is using same number of physical pixels at all DPI. If using 200% for example, everything should be double size compared to 100%. If still want to improve further, I think need to scale some of the value with the DPI.

g-maxime:
In my case (4K, 200%), scaling to the dpi ratio seems to lose all the benefits of the banner image size limitation.

Without scaling:
Without scaling

With scaling:
With scaling

cjee21:
Ya on yours, without scaling looks the right size. I don't know how the scaling/sizing works. The ad seems to always be the same size here.

QHD 100%
Screenshot 2025-09-25 150105

QHD 150%
Screenshot 2025-09-25 150141

QHD 200%
Screenshot 2025-09-25 150212

FHD 100%
Screenshot 2025-09-25 150533

JeromeMartinez:
> The ad seems to always be the same size here.

Your screenshots seem OK to me, the goal is to be not too big when the window is maximized, and the height is limited to 1/8 of the screen height (a bit bigger than a typical ad on the web, 90 pixel height for 1080 pixel screen height is 1/12).
by default the MediaInfo windows is 1/2 of the screen height and the ad height is limited to 1/4 of the window, so the ad size should be same with the default size or maximized, it seems still readable and less agressive.

JeromeMartinez:
> The ad seems to always be the same size here.

Said a bit differently: would it be good to increase the ad height if other texts are bigger? It may be too big relatively to the screen, and we would go back to the first criticisms.

cjee21:
> Your screenshots seem OK to me

If it seems okay then it's okay I think. After all, too small would be better than too big from user perspective.

cjee21:
> too small would be better than too big

I measured the ad in pixels and compared to standard web ad, on 1440p display, at 200% is is same size as web ad and 100% one is double size. So it is not too small on hi dpi compared to web ad but rather 100% one is bigger. No issues then if you decide to use fraction of screen size for sizing.

JeromeMartinez:
> So it is not too small on hi dpi compared to web ad but rather 100% one is bigger. 

Thank you for the feedback. I hesitate but I think I stay with the 1/8 max everywhere.

PR #1162, opened 2025-09-19
Android GUI: Update billing API - fixes

g-maxime:
- Fix the lifetime of the SubscriptionManager object (unable to subscribe after opening the SettingsActivity issue).
- Fix cropped legal text on small-resolution devices.

### DIFF
--- a/Source/GUI/Android/app/src/main/java/net/mediaarea/mediainfo/MediaInfoApplication.kt
+++ b/Source/GUI/Android/app/src/main/java/net/mediaarea/mediainfo/MediaInfoApplication.kt
@@ -20,4 +20,10 @@ class MediaInfoApplication : Application() {
         Lingver.init(this, InMemoryLocaleStore())
         AppCompatDelegate.setDefaultNightMode(AppCompatDelegate.MODE_NIGHT_NO)
     }
+
+    override fun onTerminate() {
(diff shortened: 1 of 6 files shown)


JeromeMartinez:
@cjee21 you may try [this build](https://mediaarea.net/download/snapshots/branches/binary/mediainfo-gui/20250919_android-billing/MediaInfo_GUI_25.07.20250919_Android.apk) for the issue between the billing pages and the other pages.

cjee21:
> @cjee21 you may try [this build](https://mediaarea.net/download/snapshots/branches/binary/mediainfo-gui/20250919_android-billing/MediaInfo_GUI_25.07.20250919_Android.apk) for the issue between the billing pages and the other pages.

Looks like it's fixed although this build appears this:

Screenshot_2025-09-22-17-06-10-027_com android vending

Now subscription page can be scrolled too.

JeromeMartinez:
@cjee21 this may be due to the .apk outside of the Play Store. if you drop your email address to info@mediaarea.net we could add you to Play Store Beta program but maybe it is not so useful because this part is tested ou our side.

cjee21:
> this may be due to the .apk outside of the Play Store

Ya maybe. Since it triggers the popup now, it should be fixed already.

PR #1164, opened 2025-09-29
Add more translation strings from the lib

JeromeMartinez:


cjee21:
```
MediaInfo_WebSite;MediaInfo web site
MediaInfoWebsite;
```

🤔 Another MediaInfo Website without translated string?

JeromeMartinez:
> 🤔 Another MediaInfo Website without translated string?

Oops... typo fixed (only one line for "both" field names)

cjee21:
Now there's an extra `MediaInfo_Website;` on the last line. 😮

JeromeMartinez:
> Now there's an extra MediaInfo_Website; on the last line. 😮

Fixed.

cjee21:
Seems there's something different with `id.csv`? Wrongly updated in one of the force push and now it is reverted but not updated like the others.

JeromeMartinez:
> Seems there's something different with `id.csv` ?

for whatever reason my script is wiping out `id.csv`  (and only it) so I don't have an automatic update for it, I'll dig in the bug when I have more free time.

PR #1165, opened 2025-09-29
MSVC PreRelease CLI for updating language files + update id.csv

cjee21:
Resolves https://github.com/MediaArea/MediaInfo/pull/1164#issuecomment-3347972158


JeromeMartinez:
I was using an old binary and maybe the bug was actually already solve in the "library" part, indeed.
Thank you for the CLI version (I was too dumb and forgot to do it).

PR #1168, opened 2025-09-30
build(cli): Adjust library name for MediaInfoLib dependency when using CMake

diizzyy:
…g CMake

CMake files for MediaInfoLib uses capitals but the CMake build script looks for library with lower-case only which causes it to fail finding it on case sensistive file systems


JeromeMartinez:
Thank you, please rebase.

diizzyy:
Actually nevermind, a lot more work is required for it to be on par with GNU Autotools.

PR #1174, opened 2025-10-06
Update Portuguese Brazil (pt-BR) translation

fulvio-alves:
I'm sending an updated translation file. I've translated some common missing strings and fixed some spelling errors.


JeromeMartinez:
Thank you.

PR #1201, opened 2025-12-05
Fix CI

cjee21:
Limit Qt version to 6.10.* because otherwise aqtinstall used by install Qt action selects version 6.11 which is not released yet (it will only feature freeze and branch from dev this week) and which it fails to install.

### DIFF
--- a/.github/workflows/MediaInfo-Qt_Checks.yml
+++ b/.github/workflows/MediaInfo-Qt_Checks.yml
@@ -16,7 +16,7 @@ on:
       - 'Source/Resource/**'
 
 env:
-  QT_VER: 6.*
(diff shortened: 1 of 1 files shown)


cjee21:
@JeromeMartinez MediaInfo wrongly detects the macOS binary as SDP, causing the other CI to fail.

Sample here: https://github.com/cjee21/MediaInfo/actions/runs/19957949937/artifacts/4774801364

JeromeMartinez:
> aInfo wrongly detects the macOS binary as SDP, causing the other CI to fail.

[Fixed](https://github.com/MediaArea/MediaInfoLib/pull/2484).

PR #1205, opened 2025-12-25
Android GUI: Dark mode rework + migrate deprecated APIs + major twoPane bugfix

cjee21:
This PR contains a suggested implementation for reworking the dark mode of the MediaInfo Android app.

Changes:
- The dark mode now defaults to follow the OS setting while overriding it to force light or dark mode regardless of OS setting remains a premium feature.
- The dark mode for report views now uses native dark mode/theme of WebView.

Reason:
- Dark mode has become an essential feature in modern apps and OS
  - Android 16 QPR2 has a feature to force dark mode on all apps
    - This means MediaInfo app will be forced to follow OS dark mode if this setting is enabled even without subscription
    - However this results in non-optimal display compared to a proper dark mode
  - There have been complaints in reviews about dark mode being a paid feature in MediaInfo
  - Even Microsoft is now implementing dark mode for the remaining Windows 11 dialogs that not yet have dark versions
- Android WebView has built-in dark mode and the HTML generated by MediaInfoLib now supports dark mode as well.
  - There is no need to manually override the colours of the report WebView anymore. Instead, using the built-in dark mode results in more consistent colours with the rest of the system.

Other changes in this PR not related to dark mode:
- Migrate ReportDetailFragment menu to MenuProvider API
  - The previously used API has been deprecated for use in fragments. It is not yet deprecated for activities so the one in main activity is not yet migrated.
- Migrate to Activity Result API for ReportDetailFragment and ReportListActivity
  - The previously used API is deprecated
  - Using new API results in less code needed too
- Migrate to ViewPager2
  - androidx.fragment.app.FragmentStatePagerAdapter is deprecated
  - https://developer.android.com/develop/ui/views/animations/vp2-migration
- Migrate AddFile functionality to Kotlin coroutines
  - AsyncTask is deprecated as well as flagged for 'StaticFieldLeak'
  - The parsing is now put in ViewModel and updating the UI is split. This is to ensure everything still works properly even if user navigates away from app or rotates screen during long parsing operations which will cause activity to be recreated.
- Fix bug with concurrent use of MediaInfo class/object.
  - The single instance of MediaInfo cannot be used for two operations concurrently and will result in corrupt parsing or reports. Even with multi instance, MediaInfo.Option is common to all instances and will cause issues.
  - This bug causes two pane mode to be unusable for multi reports. Once one report is opened, opening another will result in blank report since Android recreates fragment on return from file picker which triggers a call to MediaInfo to convert saved report while MediaInfo is currently parsing the newly opened file.
  - The same thing will happen if user taps on a previous report while the newly opened file is parsing.
  - The solution is to add an overlay for the spinner to prevent user from tapping reports during parsing as well as adding `@Synchronized` to prevent other instances of concurrent calls such as those triggered by Android.
- Use `proguard-android-optimize.txt` instead of `proguard-android.txt`
  - As recommended by Google
  - `proguard-android.txt` is deprecated and will be removed in future versions
  - `-keepclasseswithmembernames...` should not be needed since there is already something similar in `proguard-android-optimize.txt`.
- Fix English grammar in subscription condition text.
- Remove API checks that are no longer needed
  - MediaInfo app no longer supports older Android versions so these checks for older APIs are no longer necessary.
- Resolve most warnings to reduce the amount of warnings in Android Studio.

These non-dark-mode-related changes (other than the English correction) should not have any visible changes to users of the app.

There are now no more remaining warnings in Android Studio for this app. The only thing remaining is outdated androidx.room and com.android.billingclient. They cannot be updated without bumping minSdkVersion to 23 as their newer versions require that. As of December, devices running Android versions less than SDK 23 accounts for 0.4% according to Google's statistics.


JeromeMartinez:
Thank you for this proposal, I am fine with it.

cjee21:
I have resolved almost all warnings in Android Studio. Also migrated a few deprecated APIs to newer APIs. Updated the first post accordingly.

What remains is usage of deprecated AsyncTask in ReportListActivity. Also should probably migrate the main menu to MenuProvider as well since already using that for the fragment menu. These two likely requires more work since that part of codes looks complex and has high risk of breaking functionality so I did not touch them.

cjee21:
Migrated the deprecated AsyncTask to Kotlin Coroutines as well. Not as difficult as I thought.

cjee21:
May want to review the warnings in .gradle files regarding deprecated 'proguard-android.txt' and some outdated dependencies. I did not touch those as they may break stuff or cause need to increase minSdkVersion.

cjee21:
I found a bug in MediaInfo Android app in two pane mode. This bug exists both before and after changes in this PR so is not related/affected by this PR. The bug is that if a report is currently displayed in two pane mode, opening a single file will produce a corrupt report for that file. It will be blank when opened. After some debugging, I suspect it is due to Android re-creating the ReportDetailFragment for the currently displayed report upon returning from file picker triggering onCreateView again. This causes the same MediaInfoLib instance to be used from two threads in parallel, one for converting report of the currently displayed report and the other for parsing the newly opened file. This causes parsing of newly opened file to break.

I guess this bug exists since long ago and is encountered mainly by tablet users (the only phone I know that activates two pane in landscape mode is Google Pixel 9a). I also guess it is the bug encountered by this review on Google Play:
Screenshot 2026-01-07 at 01-37-01 MediaInfo - Apps on Google Play

cjee21:
> I found a bug in MediaInfo Android app in two pane mode.

I have fixed/prevented this bug as well as similar bugs from getting triggered.

> May want to review the warnings in .gradle files regarding deprecated 'proguard-android.txt'

Migrated to `proguard-android-optimize.txt` and everything still appears to be working in release mode.

> Also should probably migrate the main menu to MenuProvider as well since already using that for the fragment menu.

Decided to leave this as it is since the older API it is not deprecated for use in activities and that part of MediaInfo app is kind of complex.

I have updated the first post to reflect the current state of this PR.

JeromeMartinez:
@cjee21 thank you for your hard work, I review all your PRs soon.

JeromeMartinez:
@cjee21 thank you for the modernization.

>  Even with multi instance, MediaInfo.Option is common to all instances and will cause issues.

I need to remove all theses common options at some point, bad original design, the workaround is great.

> There are now no more remaining warnings in Android Studio for this app.

Great!

>  The only thing remaining is outdated androidx.room and com.android.billingclient. They cannot be updated without bumping minSdkVersion to 23 as their newer versions require that. As of December, devices running Android versions less than SDK 23 accounts for 0.4% according to Google's statistics.

We try to avoid to remove the support of old machines as much as possible if we don't have a big reason for doing it, and upgrading the billingclient without having to do it (no big feature, and Google does not force us to do so yet) is not enough, even if it is only 0.4%, let's keep Android 5 support for the moment, but it is not far when we remove it.

PR #1222, opened 2026-01-26
Qt GUI: Fix build with newer/custom Qt builds and bug fixes

cjee21:
It was originally written as `QFontDatabase::families()` since `QFontDatabase::QFontDatabase()` is deprecated since 6.0 and Qt docs says: "We strongly advise against using it in new code. Call the class methods as static functions instead."

It was changed to `QFontDatabase().families()` by g-maxime in https://github.com/MediaArea/MediaInfo/commit/a8ce7186702bfc80098b35b8ca43cda2c2031d66.

Now it cannot be built if using a version of Qt with disable-deprecated-up-to = 0x060000:
```
error C2280: 'QFontDatabase::QFontDatabase(void)': attempting to reference a deleted function
note: see declaration of 'QFontDatabase::QFontDatabase'
note: 'QFontDatabase::QFontDatabase(void)': function was explicitly deleted
```

So use a version check and switch between both to ensure it can be built on all versions of Qt.

Also make use of `translator->load` return value instead of `Q_UNUSED`. I should have done so from the beginning.

### DIFF
--- a/Source/GUI/Qt/easyviewwidget.cpp
+++ b/Source/GUI/Qt/easyviewwidget.cpp
@@ -26,6 +26,19 @@ EasyViewWidget::EasyViewWidget(Core* C, QWidget *parent) : QWidget(parent)
     FilePos=0;
     C->Menu_View_Easy();
 
+    fileChoice = new QComboBox(this);
+    fileChoice->setMinimumContentsLength(1);
+
+#if defined(_WIN32) && defined(WINAPI_FAMILY) && (WINAPI_FAMILY==WINAPI_FAMILY_APP) // Workaround render bug
+    QString style = "QComboBox QAbstractItemView { border: 1px solid gray }";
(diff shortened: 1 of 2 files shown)


cjee21:
@JeromeMartinez The install Qt action is failing too often so I pin it to specific publicly released version and also enable cache. In the future maybe can use MediaArea hosted Qt version if going to use a self-built static Qt.

cjee21:
Found a bug where combobox of easy view is recreated every time the filepos is changed but the old comboboxes are not deleted. Deleting a layout does not remove its members from the widget. This causes multiplying overlapping comboboxes that are visible in Windows 11 dark mode. It is not visible in light mode. ~Now fixed by deleting previous combobox before creating new one. Probably can be optimized by not re-creating comboboxes on position change.~ Fixed by only having one combobox that is re-used on refresh which should provide better performance too.

PR #1225, opened 2026-01-28
Rename MediaInfo.vcxproj to MediaInfo-CLI.vcxproj for naming consistency

cjee21:
Now we can target `MediaInfo-CLI` using MSVC2026 as well.

The only changes in build/package scripts needed for using MSVC2026 are as follows:

```diff
diff --git a/Release/Build_CLI_Windows.ps1 b/Release/Build_CLI_Windows.ps1
index 819aa556..59c08f34 100755
--- a/Release/Build_CLI_Windows.ps1
+++ b/Release/Build_CLI_Windows.ps1
@@ -15,15 +15,15 @@ $version = (Get-Content "${release_directory}\..\Project\version.txt" -Raw).Trim

 #-----------------------------------------------------------------------
 # Prepare
-Push-Location -Path "${release_directory}\..\..\zlib\contrib\vstudio\vc17"
+Push-Location -Path "${release_directory}\..\..\zlib\contrib\vstudio\vc18"
     ((Get-Content -Path zlibstat.vcxproj) -Replace 'MultiThreadedDLL','MultiThreaded') | Set-Content -Path zlibstat.vcxproj
 Pop-Location

-Push-Location -Path "${release_directory}\..\..\ZenLib\Project\MSVC2022\Library"
+Push-Location -Path "${release_directory}\..\..\ZenLib\Project\MSVC2026\Library"
     ((Get-Content -Path ZenLib.vcxproj) -Replace 'MultiThreadedDLL','MultiThreaded') | Set-Content -Path ZenLib.vcxproj
 Pop-Location

-Push-Location -Path "${release_directory}\..\..\MediaInfoLib\Project\MSVC2022"
+Push-Location -Path "${release_directory}\..\..\MediaInfoLib\Project\MSVC2026"
     ((Get-Content -Path Library\MediaInfoLib.vcxproj) -Replace 'MultiThreadedDLL','MultiThreaded') | Set-Content -Path Library\MediaInfoLib.vcxproj
     ((Get-Content -Path Dll\MediaInfoDll.vcxproj) -Replace 'MultiThreadedDLL','MultiThreaded') | Set-Content -Path Dll\MediaInfoDll.vcxproj
     ((Get-Content -Path Example\HowToUse_Dll.vcxproj) -Replace 'MultiThreadedDLL','MultiThreaded') | Set-Content -Path Example\HowToUse_Dll.vcxproj
[...]

PR #1226, opened 2026-01-28
Windows CLI: Enable using build/release scripts with VS2026

cjee21:
Implement @JeromeMartinez's idea (https://github.com/MediaArea/MediaInfo/pull/1225#issuecomment-3811303352). Now an argument can be passed after arch which can be either `MSVC2022` or `MSVC2026`. It is optional and will default to `MSVC2022` so there should be no changes needed in the build server part that calls the scripts.


JeromeMartinez:
Works well with our CI.

PR #1235, opened 2026-02-05
Windows CLI: Enable SDL Checks

cjee21:
To be consistent with MediaInfoLib.

Also make CI stricter and fix issue found due to it.

### DIFF
--- a/Source/GUI/WxWidgets/App.cpp
+++ b/Source/GUI/WxWidgets/App.cpp
@@ -27,9 +27,12 @@ IMPLEMENT_APP(App)
 bool App::OnInit()
 {
     #ifdef __WXMAC__
+        #pragma GCC diagnostic push
+        #pragma GCC diagnostic ignored "-Wdeprecated-declarations" // GetCurrentProcess is deprecated
         ProcessSerialNumber PSN;
         GetCurrentProcess(&PSN);
         TransformProcessType(&PSN, kProcessTransformToForegroundApplication);
+        #pragma GCC diagnostic pop
         setenv("LC_CTYPE", "UTF-8", 0);
     #endif //__WXMAC__
 
--- a/Source/GUI/WxWidgets/GUI_Main.cpp
+++ b/Source/GUI/WxWidgets/GUI_Main.cpp
@@ -35,6 +35,7 @@ using namespace std;
 
 //---------------------------------------------------------------------------
 //Get command line args in main()
+#undef GETCOMMANDLINE
 #ifdef UNICODE
     #ifdef _WIN32
         #include <windows.h>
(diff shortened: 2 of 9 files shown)


JeromeMartinez:
@cjee21 for rejection of deprecated APIs everywhere, change

```
        GetCurrentProcess(&PSN);
```
to
```
        
GetProcessForPID(getpid(), &psn);
```

cjee21:
> for rejection of deprecated APIs everywhere, change

`GetProcessForPID(getpid(), &psn)` is deprecated too and there are also `warning: binding dereferenced null pointer to reference has undefined behavior [-Wnull-dereference]` warnings so I just disable `-Werror` for macOS GUI part.

JeromeMartinez:
>  I just disable -Werror for macOS GUI part.

Possible to keep it and put a pragma disable only for this line? So it is explicit that it is only this line.

cjee21:
> > I just disable -Werror for macOS GUI part.
> 
> Possible to keep it and put a pragma disable only for this line? So it is explicit that it is only this line.

Then have to deal with a few of these as well:

```
../../../Source/GUI/WxWidgets/GUI_Main_Menu.cpp:560:26: warning: binding dereferenced null pointer to reference has undefined behavior [-Wnull-dereference]
  560 |         OnMenu_View_Tree(*((wxCommandEvent*)NULL));
      |                          ^~~~~~~~~~~~~~~~~~~~~~~~
```

cjee21:
@JeromeMartinez Done. Suppress only the deprecated macOS APIs section and fixed dereferenced null pointer by using temp object.

FYI some info:
Visual Studio 2026 is now available in public preview on GitHub Actions.
Visual Studio 2026 drops support of Windows versions older than Windows 10 as well as ARM (32-bit) architecture.

JeromeMartinez:
>  dereferenced null pointer by using temp object.

wouldn't "{}" or "OnMenu_View_Tree(wxCommandEvent()));" be OK while being easier to read? (afk, can not test myself).

> Visual Studio 2026 drops support of Windows versions older than Windows 10

Reason we won't use it in the next years for publishing our binaries, but it is acceptable to use it in the CI for having better compilation errors/warnings.

cjee21:
> but it is acceptable to use it in the CI for having better compilation errors/warnings

I can switch the CIs to use 2026 if you want. But right now it is in public preview and needs `runs-on: windows-2025-vs2026`. They will make it in default runners in a few months.

> wouldn't "{}" or "OnMenu_View_Tree(wxCommandEvent()));" be OK while being easier to read? (afk, can not test myself).

No idea. The function appears to want a reference to a variable/object.

AI says:
> 1. The OnMenu_View_Tree(wxCommandEvent()); Approach
>
>This is perfectly valid C++. It creates a temporary (r-value) object that lives just long enough for the function call to complete.
>
>    Pros: It’s a one-liner and eliminates the need for a named nullEvent variable.
>
>    The "Catch": Most wxWidgets event handlers are defined using a non-const reference: void OnMenu_View_Tree(wxCommandEvent& event); Standard C++ does not allow a temporary object to be passed to a non-const reference. If your compiler is strict, this will throw an error. However, many compilers (like MSVC) allow this as an extension, which is why you see it work in some projects but fail in others.
>
>2. The {} (Uniform Initialization) Approach
>
>Using OnMenu_View_Tree({}); is the C++11 (and newer) way of saying "create a default instance of whatever this function expects."
>
>    Pros: It’s the cleanest looking syntax.
>
>    The "Catch": It suffers from the same reference issue mentioned above. If the function expects a wxCommandEvent&, the compiler will likely complain that it cannot bind a temporary object to a non-const reference.

PR #1236, opened 2026-02-06
Restore ViewPager behaviour using ViewPager2

g-maxime:
Restore ViewPager behaviour using ViewPager2

Snapshot: https://mediaarea.net/download/snapshots/branches/binary/mediainfo-gui/20260401_fix-pageview2/MediaInfo_GUI_26.01.20260401_Android.apk

~~Revert to ViewPager as using ViewPager2 causes scrolling issues with WebView2~~

~~I am testing ways to maintain functional behavior with ViewPager2~~

### DIFF
--- a/Source/GUI/Android/app/src/main/java/net/mediaarea/mediainfo/NestedScrollableHost.kt
+++ b/Source/GUI/Android/app/src/main/java/net/mediaarea/mediainfo/NestedScrollableHost.kt
@@ -1,0 +1,94 @@
+package net.mediaarea.mediainfo
+
+import android.content.Context
+import android.util.AttributeSet
+import android.view.MotionEvent
+import android.view.View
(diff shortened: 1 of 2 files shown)


cjee21:
Turns out ViewPager2 does not natively support scrolling of elements inside it in the same direction.

Need to determine current scrollable state of WebView and call parent.requestDisallowInterceptTouchEvent( /* true or false */) to enable or disable ViewPager2's swiping.

https://medium.com/@nasrabadiam/support-nested-scrollable-elements-inside-viewpager2-59fa34978899

JeromeMartinez:
Works fine on my smartphone.
@cjee21 OK for you too?

cjee21:
> Works fine on my smartphone. @cjee21 OK for you too?

Seems fine in emulator, did not test on phone.

PR #1237, opened 2026-02-06
Dark Icon

panzer-ui:
Added a light/dark icon variant according to the latest macOS guidelines. The Icon must be added as AppIcon for macOS Tahoe to pick it and reflect theme changes.


cjee21:
Maybe we should do the same on Windows and invert the icon to white on dark mode for more contrast.

JeromeMartinez:
> Maybe we should do the same on Windows and invert the icon to white on dark mode for more contrast.

I would be fine with that, I use dark mode and the icon is definitely not so good, would you be motivated for doing it?

cjee21:
> would you be motivated for doing it?

I'll see how easy/difficult it is and how it looks.

JeromeMartinez:
> Maybe we should do the same on Windows and invert the icon to white on dark mode for more contrast.

@cjee21 still motivated? ;-)

cjee21:
> @cjee21 still motivated? ;-)

PR waiting for you to test isn't it?

JeromeMartinez:
> PR waiting for you to test isn't it?

Oops, I am so late, I see it.

PR #1240, opened 2026-03-09
CLI: Support Unicode paths for LogFile output

### LINKED ISSUE #1239: Windows CLI --LogFile parameter doesn't work properly with unicode chars
Hello,

Using Windows CLI version, the `--LogFile` parameter doesn't work properly with Unicode characters.

For example, this command reads the file correctly and displays the info on the console screen, but **_doesn't_** save the **log file**:
```batch
"C:\example\MediaInfo CLI\MediaInfo.exe" --LogFile="C:\example\duck🦆.m4a.MediaInfo.txt" "C:\example\duck🦆.m4a"
```

On the other hand, the following commands work properly and can be used as alternative:
```batch
"C:\example\MediaInfo CLI\MediaInfo.exe" --LogFile="C:\example\duck.m4a.MediaInfo.txt" "C:\example\duck🦆.m4a"
"C:\example\MediaInfo CLI\MediaInfo.exe" "C:\example\duck🦆.m4a" > "C:\example\duck🦆.m4a.MediaInfo.txt"
```

cjee21:
Fix #1239

It will use new filesystem features to support UTF-8 filepaths if available (C++20 or C++17) else fallback to previous behaviour. Tested on Windows with all C++ standards supported by MSVC2026.

Fixed crash that happens due to over-read if user passes `--LogFile` without equals as equals was not checked but it attempts to read after equals in the function called.

Added `File_Curl;UserAgent` option to help output as users seem to not know about it.

Update CI actions/checkout and removed C++20 flag that is no longer needed as CMake now defaults to C++20 if available which is needed for the UTF-8 filesystem features.

### DIFF
--- a/Source/CLI/Help.cpp
+++ b/Source/CLI/Help.cpp
@@ -118,6 +118,8 @@ int Help()
     TEXTOUT("--BOM");
     TEXTOUT("                    Byte order mark for UTF-8 output");
     TEXTOUT("");
+    TEXTOUT("--File_Curl;UserAgent=...");
+    TEXTOUT("                    Set the User-Agent header field in the HTTP request sent to the remote server.");
     TEXTOUT("--Ssl_CertificateFileName=...");
     TEXTOUT("                    File name of the SSL certificate.");
     TEXTOUT("                    The default format is \"PEM\" and can be changed");


cjee21:
CI requires https://github.com/MediaArea/MediaInfoLib/pull/2549 to pass.

cjee21:
Something going on, it failed at https://github.com/cjee21/MediaInfo/actions/runs/22855680275/job/66297852593 but passed over here

JeromeMartinez:
Very weird, seems a bad compilation on your run, I am for merging that PR and see later if we face again such issue.

cjee21:
> Very weird, seems a bad compilation on your run, I am for merging that PR and see later if we face again such issue.

No it is because you did not re-run all but I did. Something broke in Linux ARM64 with new C++ standard. We don't have Ubuntu ARM64 run with CMake at MediaInfoLib repo so it was missed I think.

JeromeMartinez:
> No it is because you did not re-run all but I did. Something broke in Linux ARM64 with new C++ standard.

Argh. OK, I don't merge it yet, could you try to find the issue?
"struct gain" seems fine, so would not be the source of the issue as is.

JeromeMartinez:
>  Something broke in Linux ARM64 with new C++ standard.

Looks like a false-positive from static analyzer, with `Get_Gain(0, [...]`.

cjee21:
C++23 is selected for all platforms after your suggested method. C++23 is still preview on MSVC and experimental on GCC.

PR #1242, opened 2026-03-10
Android: Statically link ZenLib to MediaInfoLib

cjee21:
So that we only have one `libmediainfo.so` in the APK and probably smaller too (I did not check). Also simplify the CMake file used for Android build and set the `cmake_minimum_required` to match MediaInfoLib. This also prevents the introduction of more `.so` files and makes it easier (only one new line of code) when we add Brotli in the future.

### DIFF
--- a/Source/GUI/Android/app/CMakeLists.txt
+++ b/Source/GUI/Android/app/CMakeLists.txt
@@ -1,7 +1,7 @@
 # For more information about using CMake with Android Studio, read the
 # documentation: https://d.android.com/studio/projects/add-native-code.html
 
-cmake_minimum_required(VERSION 3.5.0)
+cmake_minimum_required(VERSION 3.5.0...4.3.0)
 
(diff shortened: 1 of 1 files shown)


JeromeMartinez:
> So that we only have one libmediainfo.so in the APK

Right, we didn't think to that but for this case it is better.

JeromeMartinez:
[Looks fine](https://mediaarea.net/download/snapshots/branches/binary/mediainfo-gui/20260310_android/) (and 50 KiB smaller).

cjee21:
I changed to use the newly added CMake config instead of relying on cppFlags. Totally eliminate any risk of the following happening:

```
https://github.com/android/ndk/issues/1693
```

Also removed the old MPEGTS_DUPLICATE_NO and added GRAPH_NO as well as switched manual UNICODE to ENABLE_UNICODE which ZenLib controls and sorted the arguments.

PR #1250, opened 2026-04-05
Fix wx-config path when only --with-wxdir is set

mohd-akram:


Project/GNU/GUI/configure.ac
```diff
@@ -208,11 +208,7 @@ if test "$with_wxwidgets" = "yes"; then
             ])
     fi
     if test "$with_wxwidgets" = "system"; then
-        if test "${with_wx_config+set}" = set; then
-            ZenLib_Wx_Config=$with_wx_config;
-        else
-            ZenLib_Wx_Config="wx-config"
-        fi
+       ZenLib_Wx_Config=$WX_CONFIG_PATH
```
JeromeMartinez, line 211: There is a risk that it does not work in some scenarios, please keep the old behavior, et if $WX_CONFIG_PATH is defined then X else the old stuff.

JeromeMartinez (changes requested):

Project/GNU/GUI/configure.ac
mohd-akram: `with_wxwidgets` is set to `system` only if `AM_PATH_WXCONFIG` succeeds so if it enters that block then `WX_CONFIG_PATH` is set for sure.

JeromeMartinez: I don't see any problem for the moment even with old versions, so let's do with that...

JeromeMartinez (approved):

PR #1253, opened 2026-04-12
ShellExt: Optimize for early return

### LINKED ISSUE #1252: Disabling Context Menu Entry Still Loads it in Background
When you disable the option to show the context menu entry, it still loads the shell extension DLL and sparse package and all that stuff in the background, it's just that it then hides it after all that.

But it still causes the context menu to delay opening.

Please make it so when you enable/disable the context menu entry option, it properly registers/unregisters the shell extension.

------------------------------------------------------------------------

I asked Gemini about how it might be fixed and it said this. I have NOT verified it's accuracy but might be a decent starting point.

> Because `MediaInfo_PackageHelper.dll` invokes `ProvisionPackageForAllUsersAsync` and `RemovePackageAsync` (with `RemoveForAllUsers`), attempting to dynamically register or unregister the Sparse Package would silently fail without Administrator privileges, which ruins the UX of the preferences menu.
> 
> If we cannot easily unregister the package per-user without elevation, the problem goes back to the shell extension DLL itself. 
> 
> The reason the DLL causes lag even when disabled is due to a flaw in the order of operations within `dllmain.cpp`. Inside `GetState`, the code resolves shortcuts via COM interfaces (`ResolveIt`), checks if the item is a directory, extracts URLs, and reads the file into memory to parse strings—**all before it checks the registry to see if the user actually wants the context menu enabled.** By moving the registry check to the very top of `GetState`, the function can instantly return `ECS_HIDDEN` without touching the file system. Loading the DLL and reading a single registry value takes microseconds, which completely eliminates the lag you are experiencing.

cjee21:
Resolves https://github.com/MediaArea/MediaInfo/issues/1252

### DIFF
--- a/Source/WindowsShellExtension/dllmain.cpp
+++ b/Source/WindowsShellExtension/dllmain.cpp
@@ -454,8 +454,27 @@ public:
 
     IFACEMETHODIMP GetState(_In_opt_ IShellItemArray* items, _In_ BOOL okToBeSlow, _Out_ EXPCMDSTATE* cmdState) {
         // Provide state of File Explorer context menu entry
-        // Hide it if registry setting indicates that it should be disabled, else it is enabled
-        UNREFERENCED_PARAMETER(okToBeSlow);
+        // Hide it if registry setting indicates that it should be disabled or file is unsupported, else it is enabled
+
+        if (!okToBeSlow) {
+            *cmdState = ECS_DISABLED;
+            // returning E_PENDING requests that a new instance of this object be called back
+            // on a background thread so that it can do work that might be slow
+            return E_PENDING;
+        }
+
+#ifdef MEDIAINFO_QT
+        auto shellExtension{ RegGetBool(HKEY_CURRENT_USER, L"Software\\MediaArea.net\\MediaInfo", L"shellExtension") };
+        if (shellExtension.has_value() && !shellExtension.value())
+#else
+        auto ShellExtension{ RegGetDword(HKEY_CURRENT_USER, L"Software\\MediaArea\\MediaInfo", L"ShellExtension") };
+        auto ShellExtension_Folder{ RegGetDword(HKEY_CURRENT_USER, L"Software\\MediaArea\\MediaInfo", L"ShellExtension_Folder") };
+        if (ShellExtension.has_value() && !ShellExtension.value() && ShellExtension_Folder.has_value() && !ShellExtension_Folder.value())
+#endif // MEDIAINFO_QT
+        {
+            *cmdState = ECS_HIDDEN;
+            return S_OK;
+        }
 
         // Check if it is a folder or supported file extension
         bool is_folder{ false };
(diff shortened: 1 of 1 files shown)


cjee21:
Only tested Qt version locally.

cjee21:
Tested for both Qt MSIX and VCL (By replacing the shell extension DLL on a 26.01 install). No issues detected. Performance should be fast enough.

Screenshot 2026-04-13 180920

Note: the last (PID 15324) run is on a .lnk shortcut so is slightly longer.

JeromeMartinez:
I launched a snapshot.
by the way, even if it is better, there is still a latence even if we don't use it at all, wouldn't it possible to unregister the DLL in addition to this optimization? IIRC this is what I do with the mouse over (very old code, at some point I need to review that, also checking how it behaves on Win11, I never checked if the mouse over still works in it).

cjee21:
The shell extension is registered in the sparse package and the sparse package is like a fully packaged MSIX, just that the app contents are not actually in the package. So this is how Windows design works. If we try to register/unregister sparse package on setting toggle, that is like installing and uninstalling an app just from a checkbox which may cause more issues. Even on unsupported file extension, the DLL is loaded due to working around the previous bug where it does not appear on some systems if we only register the extensions we want.

There is still a bug on Windows 11 that has been around for 1~2 years where if user lanches a folder from File Explorer jumplist on taskbar, all shell extensions will be loaded and will not be unloaded (usually they are unloaded after ~10 secs of inactivity) until a logout/restart. this causes update failures for apps that have shell extensions.

Regarding the mouse over extension for MediaInfo, it worked on Windows 11 the last time I tested. I posted a screenshot previously on the issue about that not working.

JeromeMartinez:
> If we try to register/unregister sparse package

I still don't know much on that, I was expecting that the menu display is not part of a bigger thing, that they are 2 separated DLLs, so if it is the only way to do that, let's keep that.

> There is still a bug on Windows 11 [...]

Nothing we can do, no workaround, right?

> Regarding the mouse over extension for MediaInfo, it worked on Windows 11 the last time I tested. I posted a screenshot previously on the issue about that not working.

Oops, I forgot, good to know.

cjee21:
> I was expecting that the menu display is not part of a bigger thing, that they are 2 separated DLLs

The context menu one is specified in the MSIX XML manifest. That is the new way of registering shell extensions on Windows 11. No more apps manually writing to system registry keys. It is easier and prevent apps from messing up or leaving behind registry keys. For Store fully packaged apps, most of them have no way to disable the context menu entry. For ones like Mp3tag that can disable, it is the same as MediaInfo currently.

> Nothing we can do, no workaround, right?

My workaround is to not touch the jumplist at all. Nothing app developers can do (unless trying to locate and terminate the correct dllhost process when failing to update). Someone capable (famous?) can try getting Microsoft to fix it though.

cjee21:
> even if it is better, there is still a latence

The shell extension is run in separate process and should be asynchronous (it should not stop/delay/freeze the context menu from appearing and the entries usually appear one-by-one after context menu has appeared; most of them are usually instant including MediaInfo one).

JeromeMartinez:
> If we try to register/unregister sparse package

But for the VCL version, for the moment the spare package is used only for the shell extension, right?
Anyway, I understand that you want to prioritize the Qt package so sparse package for all and that you do'nt want to spend time on that.

> the entries usually appear one-by-one after context menu has appeared; 

But one person caught that the delay here and the fact that there is no immediate return :-p.

cjee21:
> But for the VCL version, for the moment the spare package is used only for the shell extension, right?
Anyway, I understand that you want to prioritize the Qt package so sparse package for all and that you do'nt want to spend time on that.

VCL just shell extension and the other effect is there is a package name shown in task manager / Process Explorer. It's not about prioritizing Qt one. Qt one is using fully packaged with own registry setting and XML manifest for now. I just want to do it the 'proper' way and prevent more issues.

> But one person caught that the delay here and the fact that there is no immediate return :-p.

Or it may be that Microsoft's bloated Notepad shell extension is taking time to load so the user checked what is going on and found our shell extension being loaded despite being disabled although that should not cause noticeable effects. I noticed usually some shell extensions appear more slowly, could be that they are bloated and slow or File Explorer is not optimized in the loading. MediaInfo shell extension is among the smallest. I wonder what Google and Microsoft put in theirs that make them large.

JeromeMartinez:
>  there is a package name shown in task manager / Process Explore

I forgot this one, thanks for reminding it. So no way to avoid the load, I need to accept it.

cjee21:
> So no way to avoid the load, I need to accept it.

Unless ask user if they want the sparse package on install. I think some apps do that. But if user wants shell extension they need to re-install.

> Or it may be that Microsoft's bloated Notepad shell extension is taking time to load

I checked and Microsoft's Notepad and Photos shell extension DLL is small so not sure why they seem to take longer than others to appear. Google's nearby share shell extension DLL is 22.2MiB! I wonder what is inside but it appears instantly as well and it looks like they have fixed their sparse package that previously contains assets.

cjee21:
Seems everything after Google's shell extension has some delay...

https://github.com/user-attachments/assets/3dcb181e-d40f-41af-937b-cae0bdf251ec

But this is only after all DLLs have been unloaded. If not it is more or less instant.

cjee21:
> > there is a package name shown in task manager / Process Explore
> 
> I forgot this one, thanks for reminding it. So no way to avoid the load, I need to accept it.

I remember another thing we use the package for... To customise taskbar icon for dark mode.

JeromeMartinez:
> I remember another thing we use the package for... To customise taskbar icon for dark mode.

OK, OK, I must accept that I need it :-p.

PR #1258, opened 2026-04-17
Android GUI: Media location permission

### LINKED ISSUE #1257: Android app no longer able to access actual file contents
All GPS related Exif data is redacted along with some XMP data. GPS coordinates for photos are all zeroes. ExifGPSVersion is also all zeroes. Location metadata of MP4 is also missing.

It is suspected to be a new behaviour introduced by the March 2026 Google Play System Update. Previously, only media picker has redacted metadata for privacy reasons but now it appears that the system file picker is also affected. MediaInfo app may need new permissions to get the actual unredacted file contents.

Opening files via Firefox in MediaInfoOnline on Android yields similar results while on Windows has full metadata so this is highly likely an Android OS behaviour and not a bug in parsing.

Reproducible on two Android devices with Android 14 and 16.

cjee21:
Resolves https://github.com/MediaArea/MediaInfo/issues/1257

### DIFF
--- a/.github/workflows/MediaInfo-Android_Checks.yml
+++ b/.github/workflows/MediaInfo-Android_Checks.yml
@@ -45,16 +45,3 @@ jobs:
           chmod +x gradlew
           ./gradlew assembleRelease --warning-mode fail
           popd
-      - name: Check alignment
-        run: |
-          BUILD_TOOLS=$(ls -d $ANDROID_HOME/build-tools/*/)
-          if [[ -n "$BUILD_TOOLS" ]]; then
-            LATEST_BUILD_TOOLS=$(echo "$BUILD_TOOLS" | sort -Vr | head -n 1)
-            export PATH="$LATEST_BUILD_TOOLS:$PATH"
-            echo "Added $LATEST_BUILD_TOOLS to PATH"
-          else
-            echo "::warning::Check alignment: Android Build Tools directory not found."
-          fi
-          curl "https://android.googlesource.com/platform/system/extras/+/main/tools/check_elf_alignment.sh?format=TEXT"| base64 --decode > check_elf_alignment.sh
-          chmod +x check_elf_alignment.sh
-          ./check_elf_alignment.sh ${{ github.workspace }}/MediaInfo/Source/GUI/Android/app/build/outputs/apk/release/app-universal-release-unsigned.apk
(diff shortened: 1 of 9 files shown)


cjee21:
Not thoroughly tested yet and not yet implemented for the file sharing from other apps path too.

cjee21:
Seems fine in my tests. I think it is better that @g-maxime implement for the handleIntent -> handleUri path since there is an existing permission prompt there that I cannot test and not sure how to integrate with the new permission. 

Unrelated note: There is a new warning `Not targeting the latest versions of Android; compatibility modes apply. Consider testing and updating this version.` in build.gradle so it seems there is a new SDK target available now but may require updating gradle plugin etc in order to increment compileSdk.

cjee21:
Works fine on my phone.

Maybe we should show some explanation to user on why the permission is needed and what happens if denied? If user denies permission then opens a file, user may wonder why still can open when permission denied or why ask permission if still can open without it. Then when user encounters coordinates all zeroes or missing location tag, user may mistakenly think their file does not have location data or has corrupt location data or that MediaInfoLib has a bug in parsing. ~It also affects some non-location related XMP data for some files, wonder why.~ XMP is another bug.

JeromeMartinez:
@cjee21 please review the [Permission request explanation for media location](https://github.com/MediaArea/MediaInfo/pull/1258/commits/05a40eb8677304db69bf19eec6107aea7063c552) commit.

JeromeMartinez:
Shouldn't the permission request related text visible only for Android version >= 16? Doable?

g-maxime:
> Shouldn't the permission request related text visible only for Android version >= 16? Doable?

The permission exists since Android 10, but seems automatically granted before Android 16...
I think it's safe to raise version checks.

cjee21:
> > Shouldn't the permission request related text visible only for Android version >= 16? Doable?
> 
> The permission exists since Android 10, but seems automatically granted before Android 16... I think it's safe to raise version checks.

No it is not automatically granted. I first noticed this issue with MediaInfo on an Android 14 phone. This phone no longer has OS updates but the behaviour changed this month which I guess is due to the March Google Play System Update.

The Android 16 AVD that's not updated also does not need the permission but Android 16 phone with March Play System Update also affected.

cjee21:
I revert back to commit 05a40eb since com.android.documentsui is a module that can be updated independently of OS since Android 10 and I am not sure if Android <12 has the March 2026 update.

For the new alert dialogs, I changed their design to use Material 3. Also added action to cancel so that user can continue to open the file if not wanting to grant permission and tapped cancel.

If user denies permission too many times or selected do not ask again then later decides to grant it, the prompt will not appear anymore so detect that and trigger another dialog which is similar to the one in preferences. In this rare case, user will have to manually open the file again after returning from system settings.

Fixed some minor text spelling as well.

g-maxime (approved):

PR #1261, opened 2026-04-21
Android GUI: Edge to edge for all Android versions

### LINKED ISSUE #1259: Android Edge-to-Edge
Playstore dashboard indicates:

Edge-to-edge may not display for all users
From Android 15, apps targeting SDK 35 will display edge-to-edge by default. Apps targeting SDK 35 should handle insets to make sure that their app displays correctly on Android 15 and later. Investigate this issue and allow time to test edge-to-edge and make the required updates. Alternatively, call enableEdgeToEdge() for Kotlin or EdgeToEdge.enable() for Java for backward compatibility.
[Learn more](https://developer.android.com/about/versions/15/behavior-changes-15#edge-to-edge)

Your app uses deprecated APIs or parameters for edge-to-edge
One or more of the APIs you use or parameters that you set for edge-to-edge and window display have been deprecated in Android 15. Your app uses the following deprecated APIs or parameters:
android.view.Window.setStatusBarColor
android.view.Window.setNavigationBarColor
These start in the following places:
com.google.android.material.datepicker.p.D
To fix this, migrate away from these APIs or parameters.
[Learn more](https://developer.android.com/about/versions/15/behavior-changes-15#edge-to-edge)

cjee21:


cjee21:
Some issues with older Android versions. The navigation bar buttons are white so not visible and the HTML goes behind the navigation bar so last lines of text are blocked.

cjee21:
Side note, the following warnings need upgrades to resolve:

A newer version of `compileSdk` than 36 is available: 37
Not targeting the latest versions of Android; compatibility modes apply. Consider testing and updating this version. Consult the `android.os.Build.VERSION_CODES` javadoc for details.
A newer version of androidx.core:core-ktx than 1.17.0 is available: 1.18.0
A newer version of androidx.room:room-runtime than 2.7.2 is available: 2.8.4
A newer version of androidx.room:room-rxjava2 than 2.7.2 is available: 2.8.4
A newer version of androidx.room:room-compiler than 2.7.2 is available: 2.8.4
A newer version of com.android.billingclient:billing than 8.0.0 is available: 8.3.0
A newer version of com.android.billingclient:billing-ktx than 8.0.0 is available: 8.3.0

JeromeMartinez:
> Some issues with older Android versions. The navigation bar buttons are white so not visible and the HTML goes behind the navigation bar so last lines of text are blocked.

So it would be better not to do it, right?

> A newer version of compileSdk than 36 is available: 37

Motivated? ;-)

> A newer version of com.android.billingclient:billing than 8.0.0 is available: 8.3.0

IIRC this is breaking some stuff, so maybe better not to change that quickly.

cjee21:
> So it would be better not to do it, right?

The overlap with HTML was probably why I did not use `enableEdgeToEdge` previously when implementing edge to edge. However the white buttons not being very visible already happens even without this `enableEdgeToEdge` change.

> Motivated? ;-)

Yes but will need to see if it requires other changes and how big a change.

JeromeMartinez:
> The overlap with HTML was probably why I did not use enableEdgeToEdge previously when implementing edge to edge. 

I am reluctant to merge the PR for this regression, people don't complain about the  white buttons but I am still reluctant to break support of old Android without a good reason.

> Yes but will need to see if it requires other changes and how big a change.

You do it better than us, please keep doing it.

BTW, com.android.billingclient:billing 8.3.0 still compatible with Android 5 so we could do an upgrade at some point too.
Not sure about when we do changes, it is still some QA to do without a precise need.

cjee21:
> I am reluctant to merge the PR for this regression, people don't complain about the white buttons but I am still reluctant to break support of old Android without a good reason.

I do not know why different behaviour with older versions. `enableEdgeToEdge` is supposed to bring edge to edge to older Android versions for consistency with Android 15+.

cjee21:
It works fine on physical Android 14 device with latest updates. Emulator with old Android 7.1 (without app updates) is the one with issues. With `enableEdgeToEdge` now the Android 14 device looks same as Android 15+ with no black bars on sides in landscape orientation.

JeromeMartinez:
> With enableEdgeToEdge now the Android 14 device looks same as Android 15+ with no black bars on sides in landscape orientation.

I am slowly convinced that small bugs with old devices are maybe not a big issue vs advantage with new devices.
A last hope: no workaround in mind for the bug you find on old devices?

cjee21:
> no workaround in mind for the bug you find on old devices?

So far nothing I tried works. The setOnApplyWindowInsetsListener does not fire on Android 8.

Another workaround (not really a workaround but maybe the actual recommended way) is to let the HTML and other pages actually go behind the nav bar on all Android versions but offset the content so that they can scroll above the nav bar when reaching the end. Not sure how to implement though.

cjee21:
Okay found the cause and solution. Will make changes for the other pages too then test and force push.

cjee21:
Seems fine in Android 8.1 and 16 emulator. Now just need test on physical devices to make sure no more issues.

cjee21:
Okay, I find no issues in my tests.

g-maxime (approved):

PR #1267, opened 2026-04-28
Android: Only export JNI_OnLoad in libmediainfo.so

cjee21:
By default, NDK linker follows ELF default of making every symbol visible. By only exporting the single symbol we actually need (JNI_OnLoad), the size of the .so file and APK can be further reduced due to removal of symbol names and unreachable code. It also improves library load performance as there are less symbols. Also prevents conflicts with other .so files that may be added to app in future due to symbols from statically linked libc++ that may be visible otherwise.

### DIFF
--- a/.github/workflows/MediaInfo-Android_Checks.yml
+++ b/.github/workflows/MediaInfo-Android_Checks.yml
@@ -38,9 +38,13 @@ jobs:
         uses: actions/checkout@v6
         with:
           path: MediaInfo
+      - name: Setup Java
+        uses: actions/setup-java@v5
+        with:
+          distribution: 'jetbrains'
+          java-version: ${{ env.JAVA_VER }}
       - name: Build APKs
         run: |
-          export JAVA_HOME="$JAVA_HOME_${{ env.JAVA_VER }}_X64"
           pushd ${{ github.workspace }}/MediaInfo/Source/GUI/Android
           chmod +x gradlew
           ./gradlew assembleRelease --warning-mode fail
--- a/Source/GUI/Android/app/CMakeLists.txt
+++ b/Source/GUI/Android/app/CMakeLists.txt
@@ -12,6 +12,20 @@ set(BUILD_ZENLIB ON CACHE BOOL "Build bundled ZenLib" FORCE)
 
 add_subdirectory(../../../../../MediaInfoLib/Project/CMake ${CMAKE_CURRENT_BINARY_DIR}/MediaInfoLib)
 
+target_link_options(mediainfo
+        PRIVATE
+        -Wl,--version-script,${CMAKE_CURRENT_SOURCE_DIR}/libmediainfo.map.txt
+        # This causes the linker to emit an error when a version script names a
+        # symbol that is not found, rather than silently ignoring that line.
+        -Wl,--no-undefined-version
+)
+
+# Without this, changes to the version script will not cause the library to relink.
+set_target_properties(mediainfo
+        PROPERTIES
+        LINK_DEPENDS ${CMAKE_CURRENT_SOURCE_DIR}/libmediainfo.map.txt
+)
+
 # disable warning in aes_via_ace.h for x86 arch
 if(NOT "${CMAKE_VERSION}" VERSION_LESS "3.18")
     set_property(
--- a/Source/GUI/Android/app/libmediainfo.map.txt
+++ b/Source/GUI/Android/app/libmediainfo.map.txt
@@ -1,0 +1,14 @@
+# The name used here also doesn't matter. This is the name of the "version"
+# which matters when the version script is actually used to create multiple
+# versions of the same symbol, but that's not what we're doing.
+LIBMEDIAINFO {
+  global:
+    # Every symbol named in this section will have "default" (that is, public)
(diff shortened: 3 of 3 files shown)


cjee21:
Size comparison of this PR vs current master.

Screenshot 2026-04-28 195956

cjee21:
If really want to optimize size, there is duplicate detected by SizeBench on the Windows build

Screenshot 2026-04-30 005711

JeromeMartinez:
> If really want to optimize size

We do :).
(and this one is easy)

cjee21:
> > If really want to optimize size
> 
> We do :)

Then this is the app that may help: https://apps.microsoft.com/detail/9ndf4n1wg7d6?hl=en-US&gl=US

cjee21:
Unrelated info: MediaInfo app does work on Android 16 desktop mode.

screen-20260504-185940-1777892077923_exported_217058~2

g-maxime (approved):
[Snapshot](https://mediaarea.net/download/snapshots/branches/binary/mediainfo-gui/20260504_android-export/)

JeromeMartinez:
@cjee21 impact of your work on release size in production :).
(and this despite all the new formats support)

image

cjee21:
Interesting... This time the split APKs delivered by Google Play have different and higher minSDKs. The one I got on my phone is minSDK 32 (12L).

From user submissions at APKmirror:
Screenshot 2026-05-13 at 00-11-04 MediaInfo 26 05 (120-480dpi) (Android 12L ) APK Download by MediaArea net - APKMirror

JeromeMartinez:
> Interesting... This time the split APKs delivered by Google Play have different and higher minSDKs. The one I got on my phone is minSDK 32 (12L).

Maybe optimized apk? Play Store admin UI shows "API levels 21+" and Paly Store user UI shows "Android 5.0 or more" for 26.05.

> From user submissions at APKmirror

do you have an idea why DPI support is different? Google internal stuff?

JeromeMartinez:
Would it be worth it to provide on our website the "Signed, universal APK" from Google so third party uploaders don't manage several versions from elsewhere, or will they do that anyway?

JeromeMartinez:
> do you have an idea why DPI support is different? Google internal stuff?

I answer to myself:
12 generated split APKs for supported devices with API level 21+
1 master split APK: master
4 ABI split APKs: arm64-v8a, armeabi-v7a, x86, x86_64
7 density split APKs: hdpi, ldpi, mdpi, tvdpi, xhdpi, xxhdpi, xxxhdpi

So I guess the DPI depends on the density split APK found by the user, and I guess that the Android version comes from when the density was available or similar.

cjee21:
> I guess that the Android version comes from when the density was available or similar.

No idea. Previous versions show the actual minSDK we put even with the splits. All splits appear to have same minSDK previously. Now I see that Google's apps that were updated recently all have minSDK 32 on my phone too.

> Would it be worth it to provide on our website the "Signed, universal APK" from Google so third party uploaders don't manage several versions from elsewhere, or will they do that anyway?

I don't know. Looking at the submissions, some versions contain the non-split APK while some contain splits.

cjee21:
Hmm... Maybe the binary can be even smaller without using wide chars (building without UNICODE)? But is that non Unicode path well tested?

JeromeMartinez:
> Hmm... Maybe the binary can be even smaller without using wide chars (building without UNICODE)? But is that non Unicode path well tested?

It is planned to remove UNICODE compilation directive everywhere but there is a lot of QA to do and also to handle correctly the ABI.

cjee21:
Well, I just tried. It is smaller...

Screenshot 2026-05-13 022028

and it can build and run... BUT it is buggy. CJK characters from Exif/XMP data are wrong. Writing hardware is duplicated many times.

The MediaInfoOnline is also buggy with the same fields for the same file but with different output than Android.

JeromeMartinez:
> It is smaller...

Less than I hoped. But the move would be also for RAM usage and small performance improvement.

>  CJK characters from Exif/XMP data are wrong. Writing hardware is duplicated many times.

Not surprised. There are some conversions from file code page to UTF8 to do.

> The MediaInfoOnline is also buggy with the same fields for the same file but with different output than Android.

another issue (reading from MediaInfo XML), also on my todo-list.

cjee21:
> > CJK characters from Exif/XMP data are wrong. Writing hardware is duplicated many times.
> 
> Not surprised. There are some conversions from file code page to UTF8 to do.

@JeromeMartinez Managed to get non-UNICODE UTF-8 working 'properly' (no issues that I noticed in short test) on Windows CLI and Android.

"CJK characters from Exif/XMP data are wrong" part is due to `IsBadChar` function thinking the valid UTF-8 chars are bad and replacing some bytes with '?', corrupting the UTF-8 encoding.

"Writing hardware is duplicated many times" part is due to `Fill` having two signatures for Ztring and string. When switch to non-UNICODE, it switch to the other signature then true becomes for `Utf8` instead of `Replace`. So it fills duplicate instead of replacing in the curate codes. No idea why Android repeats more times than Windows though, maybe to do with XML import.

The MediaInfoOnline issues are likely due to these as well since the symptoms are similar. I assume that one is using UTF-8 / non-UNICODE.

PR #1270, opened 2026-05-05
Android GUI: Improve layout and accessibility

cjee21:
- Make report list able to scroll content above the floating action button (FAB) so that content is not blocked by it.
- Make FAB stay within report list in two pane mode since it belongs there and so it doesn't block the report.
- Improve colour contrast of report list buttons for accessibility and to remove IDE warnings.
- Increase touch size of report list items for accessibility and to remove IDE errors. [48dp is minimum for touch target size](https://support.google.com/accessibility/android/answer/7101858?hl=en).
- Remove redundant XML attributes that IDE says is redundant for MediaInfo's current targetSDK.

Screenshot_20260505_180234

Screenshot_20260505_180347

### DIFF
--- a/Source/GUI/Android/app/src/main/java/net/mediaarea/mediainfo/ReportListActivity.kt
+++ b/Source/GUI/Android/app/src/main/java/net/mediaarea/mediainfo/ReportListActivity.kt
@@ -13,6 +13,7 @@ import androidx.activity.result.contract.ActivityResultContracts
 import androidx.annotation.RequiresApi
 import androidx.appcompat.app.AppCompatActivity
 import androidx.appcompat.app.AppCompatDelegate
+import androidx.coordinatorlayout.widget.CoordinatorLayout
 import androidx.core.app.ActivityCompat
 import androidx.core.content.edit
 import androidx.core.net.toUri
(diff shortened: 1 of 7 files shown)


cjee21:
There seems to be performance issues with current implementation of report list according to AI. IIRC while testing with hundreds of items, scrolling performance is not great. I did not attempt to fix this as it seems to require big changes.

Screenshot 2026-05-05 164157

JeromeMartinez:
> There seems to be performance issues with current implementation of report list according to AI. IIRC while testing with hundreds of items, scrolling performance is not great. I did not attempt to fix this as it seems to require big changes.

Definitely not part of this PR but I am interested in a fix for that if you are motivated.
Reports are only few KB big, so should not use a lot of RAM to display, especially because there is only one visible report.

cjee21:
> Definitely not part of this PR but I am interested in a fix for that if you are motivated.

I looked into it earlier but it seems to require big change since it is recommended to completely remove the NestedScrollView. I am also not very familiar with the part handling the list. Therefore, not very motivated especially that I do not usually keep reports stored so technically does not affect my use.

> Reports are only few KB big, so should not use a lot of RAM to display, especially because there is only one visible report.

It is not about the report size or its display. It is the list of reports. If user has many reports, the list is very long and right now it appears to be loading/rendering the entire long list and not just the items that are currently in view on screen.

g-maxime (approved):

g-maxime:
[Snapshot](https://mediaarea.net/download/snapshots/branches/binary/mediainfo-gui/20260505_android/MediaInfo_GUI_26.01.20260505_Android.apk)

PR #1271, opened 2026-05-05
Android GUI: Report list improvements

cjee21:
Get the RecyclerView out of NestedScrollView so that it can work properly and should have better performance and efficiency. Also show scrollbar so that users have an idea how long their report list is.

FAB positioning is now done with XML since we have RecyclerView with a fixed ID that can anchor to. Also put it in a FrameLayout so that it has similar margins when in two pane mode.


g-maxime (approved):
[Snapshot](https://mediaarea.net/download/snapshots/branches/binary/mediainfo-gui/20260505_Android-2/MediaInfo_GUI_26.01.20260505_Android.apk)

PR #1276, opened 2026-05-13
Windows: Enable Segment Heap + Unify resource file

cjee21:
- Opt-in to using Segment Heap for Windows CLI and Qt GUI
   As described by Microsoft in latest MSVC changelog:
   > New MSBuild and CMake C++ projects now use Segment Heap by default, helping applications take advantage of modern Windows memory management and security features.
   >
   > Visual Studio now enables Segment Heap by default for new C++ projects. Segment Heap is a modern Windows heap implementation that improves performance, scalability, and security compared to the legacy NT heap. This change helps applications adopt these benefits and prepares them for future platform capabilities.
   >
   > You can configure this behavior in project properties under Manifest Tool → Input and Output. Existing projects are not modified automatically but can opt in using the same settings. 

- Unify resource file
   I just noticed that the MSVC2026 binary has version 25.10. This is because the resource file is one copy per MSVC version and stored together with project file and version update script was not updated to handle the new one. So now follow MediaInfoLib where there is only one resource file and stored in sources directory. This also makes CMake one not dependent on presence of MSVC2022 files. @g-maxime need to update version script to update the new location when this PR is merged.


JeromeMartinez (dismissed):
Segment Heap is not compatible with Windows 7 (AI says that behavior is random, sometimes crashes), and we still support it.
MSVC 2026 removed the support of Windows 7 so no problem there, but MSVC 2022 still has the support of Windows 7 and we currently use that in production for this reason.

Please remove this feature from MSVC 2022 project.
Then it will be ready for when we stop to support Windows 7 (we'll move to MSVC 2026 when it is the case).

cjee21:
> Segment Heap is not compatible with Windows 7

It is supported on Windows 10 2004 onwards. Since it is only a flag in manifest of exe, older Windows should just ignore it.

cjee21:
> AI says that behavior is random, sometimes crashes

The AI I use says fine which is what I thought as well.

Screenshot 2026-05-13 at 16-42-21 Segment Heap Backwards Compatibility Explained - Google Gemini

cjee21:
Seems packaged apps already default to Segment Heap.

Checking MediaInfo Qt installed from MSIX package:

```
0:006> !heap
        Heap Address      NT/Segment Heap

         134d5000000         Segment Heap
```

PR #1279, opened 2026-05-14
Android GUI: Changes for minSDK >= 23

cjee21:
This PR contains updates and improvements that can be made when minSDK is increased to 23.

- Update dependencies including those that require minSDK 23
- Migrate to Material 3 Expressive theme
  - Material 1.14.0 stable is released with Expressive now being stable
  - This changes look of loading circle and maybe some other elements
  - Change report list to new look with cards similar to Gmail etc. with currently selected report highlighted in twopane mode
  - Preferences is not migrated to card look as it seems difficult
  - Floating Action Button is now extended type that can expand and shrink on scroll
- Also update activities to have app bar the same colour as background and only change to darker when scrolled
- Add libcurl support and support passing weblinks to the app via sharing from other apps (require changes to MediaInfoLib's CMake to work)

Screenshot_20260516_004058
Screenshot_20260516_003549

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -68,6 +68,7 @@ build/
 taskfile.yaml
 
 # Android app directories
+**/.cxx/*
 /Source/GUI/Android/.gradle
 /Source/GUI/Android/.idea
 /Source/GUI/Android/app/build
--- a/Source/GUI/Android/app/CMakeLists.txt
+++ b/Source/GUI/Android/app/CMakeLists.txt
@@ -10,6 +10,8 @@ set(BUILD_SHARED_LIBS ON CACHE BOOL "Force shared library build" FORCE)
 set(LARGE_FILES OFF CACHE BOOL "Enable large files support" FORCE)
 set(BUILD_ZENLIB ON CACHE BOOL "Build bundled ZenLib" FORCE)
 
+find_package(curl REQUIRED CONFIG)
+
 add_subdirectory(../../../../../MediaInfoLib/Project/CMake ${CMAKE_CURRENT_BINARY_DIR}/MediaInfoLib)
 
 target_link_options(mediainfo
--- a/Source/GUI/Android/app/build.gradle
+++ b/Source/GUI/Android/app/build.gradle
@@ -32,20 +32,18 @@ android {
                         "-DMEDIAINFO_DUPLICATE=OFF",
                         "-DMEDIAINFO_DVDIF_ANALYZE=OFF",
                         "-DMEDIAINFO_EVENTS=OFF",
+                        "-DMEDIAINFO_FILE=OFF",
                         "-DMEDIAINFO_FILTER=OFF",
                         "-DMEDIAINFO_FIXITY=OFF",
                         "-DMEDIAINFO_GRAPH=OFF",
                         "-DMEDIAINFO_IBI=OFF",
-                        "-DMEDIAINFO_LIBCURL=OFF",
                         "-DMEDIAINFO_LIBMMS=OFF",
                         "-DMEDIAINFO_MACROBLOCKS=OFF",
                         "-DMEDIAINFO_MD5=OFF",
                         "-DMEDIAINFO_NEXTPACKET=OFF",
-                        "-DMEDIAINFO_READER=OFF",
                         "-DMEDIAINFO_READTHREAD=OFF",
                         "-DMEDIAINFO_REFERENCES=OFF",
                         "-DMEDIAINFO_SHA1=OFF",
-                        "-DMEDIAINFO_SHA2=OFF",
                         "-DMEDIAINFO_TRACE=OFF",
                         "-DMEDIAINFO_TRACE_FFV1CONTENT=OFF"
             }
@@ -76,6 +74,9 @@ android {
             shrinkResources = true
             proguardFiles getDefaultProguardFile('proguard-android-optimize.txt'), 'proguard-rules.pro'
         }
+    }
+    buildFeatures {
+        prefab = true
     }
     externalNativeBuild {
         cmake {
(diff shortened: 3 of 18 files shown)


JeromeMartinez:
Great!
As we are there, would it be possible to have the "+" icon bigger in portrait mode?

> Preferences is not migrated to card look as it seems difficult

It would be great to have a simpler code here and all with the same style, if you find a way... I know, I ask a lot.

cjee21:
And I thought `Android GUI: Highlight currently opened report` is working well but crashes when list is cleared... 

> As we are there, would it be possible to have the "+" icon bigger in portrait mode?

I was thinking making it expand and shrink on scroll where it shows some text next to it. Like the buttons in Google's apps.

cjee21:
> And I thought Android GUI: Highlight currently opened report is working well but crashes when list is cleared...

It works great now and with amazing and smooth animations when list is updated. Also maintains position of list on navigation. Previously it wasn't very efficient with adapter being recreated on navigation/add and not letting native code handle the list.

cjee21:
Looks like that's all for now. Updated screenshots above. Unable to use native shape/color changing (https://github.com/material-components/material-components-android/blob/master/docs/components/List.md#listitemcardview-attributes) for currently selected item; the colour manually toggled with XML and shape no change. Preferences seems difficult as PreferenceScreen is not updated to new style so may need to migrate the entire thing to ListItemCardView to make it look like new Android system settings style.

cjee21:
Since @JeromeMartinez said "Note: it would be also nice to have that with the Android UI", added the initial libcurl support to the Android GUI. At the moment can only share from other apps, somewhat similar to Windows GUIs where it can be passed via command line but no GUI for inputting URLs yet.

[Screen_recording_20260518_192357.webm](https://github.com/user-attachments/assets/375e2d0b-6d89-4189-8162-cfb0cc5225c0)

Not thoroughly tested and have not checked size impact yet.

> [!NOTE]
> Requires changes to MediaInfoLib's CMake (that are in the android-lib PR) to detect and link with libcurl

Side note: enabling libcurl require enabling reader to work then after that requires enabling sha2 which then requires enabling md5 in order to build successfully with `Reader_libcurl` enabled.

JeromeMartinez:
>  added the initial libcurl support to the Android GUI.

Thank you!

> At the moment can only share from other apps

Not a bid deal.

> have not checked size impact yet.

this would be a concern, but checking libcurl size it seems acceptable (0.5 MB compressed per arch?). I like to have things as small as possible but it is not like if MediaInfo is huge.
An alternative could be to use Dynamic Feature Modules for libcurl packaging (note: I have also in mind to have translations as Dynamic Feature Modules in order to reduce delivery size)

> Side note: enabling libcurl require enabling reader to work then after that requires enabling sha2 which then requires enabling md5 in order to build successfully with curl_reader enabled.

Acceptable, but you can add MEDIAINFO_FILE_NO and MEDIAINFO_DIRECTORY_NO, not needed.

> implementation "io.github.vvb2060.ndk:curl:8.18.0"

I have a concern here, it is outdated by few months (2 new versions in the meanwhile), but at longer term we could use our own build for also being safer.

cjee21:
> > implementation "io.github.vvb2060.ndk:curl:8.18.0"
> 
> I have a concern here, it is outdated by few months (2 new versions in the meanwhile), but at longer term we could use our own build for also being safer.

I cannot find a more official/reliable curl for Android. Anyway, all the build scripts are on GitHub for this one so we probably can update it and build it ourselves for more reliability, control and security.

> it is outdated by few months

Isn't the libcurl for the MediaInfo Windows releases way older than this? 😅

JeromeMartinez:
> Isn't the libcurl for the MediaInfo Windows releases way older than this? 😅

Something I am not proud about, I need to update that more often.

cjee21:
> Something I am not proud about, I need to update that more often.

Since it is from vcpkg, maybe can script an automated build whenever new version is available.

Same with Graphviz. Can build including ARM64 from vcpkg.

JeromeMartinez:
> Since it is from vcpkg, maybe can script an automated build whenever new version is available.

I could accept such automation ;-).

cjee21:
Disabled FILE (DIRECTORY was already disabled) and also managed to exclude MD5 which requires fix in MediaInfoLib so made a new PR there.

cjee21:
> but checking libcurl size it seems acceptable (0.5 MB compressed per arch?)

Seems libmediainfo.so increased 2MiB uncompressed.

By increasing minSDK to 23, native libraries are now uncompressed in APK to allow direct loading without extraction. This results in smaller install size due to no duplicate from extraction. Play Store will compress the transfer so download size is fine. Just that APK file size will be larger.

cjee21:
> > Since it is from vcpkg, maybe can script an automated build whenever new version is available.
> 
> I could accept such automation ;-).

@JeromeMartinez Example of automated libcurl build: https://github.com/cjee21/libcurl-Windows. Can integrate this with one of MediaArea's repository. Then in build/packaging CI, can fetch latest from a fixed URL like mentioned in the readme.

JeromeMartinez:
>  Then in build/packaging CI, can fetch latest from a fixed URL like mentioned in the readme

https://github.com/MediaArea/libcurl-Windows
@g-maxime please automate, including test of non regression.

cjee21:
There is a bug, opening a url after a file, MediaInfo shows the previous filename as complete name instead of the url.

Update: Solved by resetting File_FileName before opening urls.

cjee21:
Screenshot_20260615-132005~2

Google putting Views in maintenance mode and going all Jetpack Compose now. In the long term we need a full rewrite of the GUI.

PR #1294, opened 2026-06-05
Qt MSIX: Rework build/release scripts, support MSVC2026/ARM64

cjee21:
Migrate Qt MSIX build/release script to PowerShell and split build and release stage in line with the other updated build/release scripts.

Enable selection between MSVC2022/2026 and x64/ARM64.

Also fix building Qt GUI variant of shell extension for ARM64.


cjee21:
Test build: https://github.com/cjee21/MediaInfo/actions/runs/27015621431

This test build contains Qt GUI for both Windows x64 and ARM64. Built using the scripts in this PR on x64 build host (ARM64 is cross-compiled). Includes libcurl 8.20.0 and Graphviz 15.0.0 from automated GItHub Actions build. All binaries are fully native x64/ARM64 without any ARM64EC except ffmpeg.exe that is x64 on both archs due to lack of ARM64 FFmpeg plugin.

cjee21:
Note: It now also produces a 7z archive of Qt GUI besides the MSIX.

JeromeMartinez:
> without any ARM64EC except ffmpeg.exe that is x64 on both archs due to lack of ARM64 FFmpeg plugin.

Looks like we need to have that now.

PR #1296, opened 2026-06-10
CI: Migrate to VS2026

cjee21:


cjee21:
Just realized Windows ARM image does not have VS2026.

JeromeMartinez:
> Just realized Windows ARM image does not have VS2026.

Can you adapt the PR for using VS2022 or VS2026 depending on the arch?

PR #1297, opened 2026-06-12
CI: VS2026 on ARM + Ubuntu 26.04 + macOS 26

cjee21:


JeromeMartinez:
@cjee21 something does not work with macos26 (the MIL part is fixed now, this is something else).

cjee21:
> @cjee21 something does not work with macos26 (the MIL part is fixed now, this is something else).

Ya. Don't know why Linux and macOS often different even using same configure. Will look tomorrow. Have some other warnings that need cleaning up too.

cjee21:
So previously, CI was building libmediainfo with dynamic libzen but both libmediainfo and libzen are static linked to CLI. In Ubuntu 26.04, CLI cannot find dynamic libzen at runtime like this. If make libmediainfo use static libzen too, it works on Ubuntu 26.04 but macOS 26 has 'libzen.a is not a Mach-O' during linking of libmediainfo then breaks when linking CLI with missing libmediainfo symbols.

cjee21:
So, keep the original 'static linking that's not fully static' and set path so that both Ubuntu and macOS can find the dynamic libraries at runtime.

PR #1302, opened 2026-07-26
Android: Fix geolocation help dialog display logic

### LINKED ISSUE #1301: Android Geolocation data access multiple requests
From a Google Play comment:

"Maybe add a option to stop the nagging about "Geolocation data access" without having to give you the information?? There is not a single reason you need to ask for that information everytime I open a audio file."

or

"Is it possible that every time you try to activate GPS, you have to be asked to do so?"

If a user does not like to have all information, why not, maybe we accept the "no" forever after 1 or 2 attempts and we have the option only in the preferences.

g-maxime:
Fixes #1301 

[Snapshot](https://mediaarea.net/download/snapshots/branches/binary/mediainfo-gui/20260725_permission-request/MediaInfo_GUI_26.05.20260725_Android.apk)


cjee21 (approved):
Works fine. Seems some users do not really understand Android's permission system and may 'complain' that MediaInfo 'bypasses' permission by still being able to access media after denying permission (there is a review on Google Play about no permissions asked previously) but there is no way to fully satisfy everyone.

